PR #12, opened 2023-04-25
Changes to support multiple test schedules

pyth0n1c:
Allow different schedules (with names) to be declared in
contentctl.yml.  Then, using the field named scheduling in
a detection, choose the schedule for this detection to run.


pyth0n1c:
The codebase has diverged significantly from this structure.
This functionality will now be implemented by adding new deployments and referencing a different `type:` in a detection

PR #23, opened 2023-06-22
Add support for auth tokens and env vars

linuxdaemon:
Load username, password, auth token, and server IP from environment variables if not set in config.

Also add support for the auth tokens.


Phenomanan (approved):

josehelps:
Hey I digged into adding this to the api_deploy action but it seems incomplete, and since we have made some major changes to the tool we should revisit this as a proper feature for customers soon. Closing this PR for now since we might re-write the hole thing. I created this issue to track the feature: https://github.com/splunk/contentctl/issues/173

PR #26, opened 2023-07-11
Implementing error handling for 409 errors indicating the search alre…

benranft12:
…ady exists

### DIFF
--- a/contentctl/actions/deploy.py
+++ b/contentctl/actions/deploy.py
@@ -3,6 +3,7 @@ import sys
 import json
 import requests
 from requests.auth import HTTPBasicAuth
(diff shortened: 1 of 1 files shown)


pyth0n1c:
Hi @benranft12 - in release v4.1.0, we have actually removed the API Deploy functionality for now: https://github.com/splunk/contentctl/releases/tag/v4.1.0

 This is because we are currently unable to create content besides searches and macros, most notably lookups (and other content like stories, etc).
Since we can't create lookups, which may be critical for a search to work as anticipated, we have removed the functionality for now.
As such, I am going to close out this PR.

PR #36, opened 2023-08-09
Allow absent tests for experimental detections

linuxdaemon:
This is consistent with the validation check
that allows no tests if status=experimental


pyth0n1c (approved):
Had to make some minor changes to use the value of the enum and ALWAYS run validator for tests, not just when the field is present (which is the default behavior).

PR #38, opened 2023-08-10
Ensure detections show up in the ES app

linuxdaemon:
This ensures that detections show up as
content in the Enterprise Security UI in Splunk

### DIFF
--- a/contentctl/actions/api_deploy.py
+++ b/contentctl/actions/api_deploy.py
@@ -47,7 +47,7 @@ class API_Deploy:
                 "username": target.username,
                 "password": target.password,
                 "owner": "nobody",
-                "app": "search",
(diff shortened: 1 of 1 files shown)


pyth0n1c:
@linuxdaemon , great catch! Instead of hardcoding this, do you the "app" field should be specifiable in the contentctl.yml file (or similarly on the command line) with a _default_ of SplunkEnterpriseSecuritySuite ?  
I anticipate there will be users of contentctl who do not have Enterprise Security and might want to create them in a different app (although I think this will be a less common use case).
Any thoughts?

linuxdaemon:
Yeah that makes sense to me, I'll add that to my todo

josehelps (approved):

josehelps:
Lets build the config changes into another PR, for now this is perfect

PR #42, opened 2023-08-14
Handle stopped containers in testing

linuxdaemon:
This means a failed container will no longer hang the test command


pyth0n1c (approved):
This looks great, but I will test a bit more exhaustively under a few different container-stopping scenarios to make sure we catch them all :)

PR #43, opened 2023-08-14
Generate a macro filw for each detection

linuxdaemon:
This allows easy management of the filters alongside detections in a
Detection-as-Code repo

### DIFF
--- a/contentctl/output/conf_output.py
+++ b/contentctl/output/conf_output.py
@@ -54,10 +54,6 @@ class ConfOutput:
             ConfWriter.writeConfFile(self.output_path/'default/analyticstories.conf',
                 'analyticstories_detections.j2',
                 self.config, objects)
-
-            ConfWriter.writeConfFile(self.output_path/'default/macros.conf',
-                'macros_detections.j2', 
-                self.config, objects)
         
         elif type == SecurityContentType.stories:
             ConfWriter.writeConfFile(self.output_path/'default/analyticstories.conf', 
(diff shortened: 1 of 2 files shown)


pyth0n1c:
I _really_ like the idea of this PR.  I have two potential questions for @linuxdaemon and hope to get your thoughts:

Instead of writing these macros to the macros/ directory, should we:
1) write them to macros/filters (a new directory) so that we can group them all together?  Or;
2) Should we consider making the filter macro definable as PART of the detection YML itself?  I'm sure that most (but not all) users will want an empty filter macro just like we ship in ESCU. But for users who want to overload this macro, it might make more sense having it located in the same file as the relevant detection.

2) We now have enhanced validation that ensure all macros used in a search actually exist as parsed macros in your repo.  I will need to add a bit of code to make sure this check passes with the new macro objects being created.

linuxdaemon:
yeah this sounds good to me, I'll probably add a config flag to determine how it manages the macros, just for compatibility (since my team has been using my fork)

pyth0n1c:
I know this change has been pending for a long time, but we have a new branch (which has a HUGE number of changes) that should address this.  It gives the following possibilities:
1.  If the filter macro exists and is defined in the `macros/` folder, then use it.  This way, you can create the filter macro and customize it.
2.  If the filter macro does NOT exist in the `macros/` folder, then create the macro at runtime.
I believe this functionality should give the same behavior proposed in this PR.  As such, I am going to close it pending the release of the linked branch:
https://github.com/splunk/contentctl/blob/a6bc860c18f54a7568494f929c707d889269c8b2/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L136

PR #44, opened 2023-08-14
Update new content generator with new formats

linuxdaemon:
The schema for the yaml files changed, this updates the new_content_generator to match, producing mostly valid detections for users.


pyth0n1c (approved):
Tested and working! Thank you for this PR.  content created in this way still does not pass validate immediately, but this is _by design_.  Fields like "risk_score" must still be entered manually.

PR #46, opened 2023-08-15
fix_no_lookup_folder_in_app

kyokkygo:
Hi!  ก₍⸍⸌̣ʷ̣̫⸍̣⸌₎ค

When I build an application with lookups I get an error that the lookups folder is not visible in dict/app. So my .csv files turn into one file called lookups and no extension.

I fixed this by creating a search folder existence check. If not, create dir.

### DIFF
--- a/contentctl/output/conf_output.py
+++ b/contentctl/output/conf_output.py
@@ -115,9 +115,14 @@ class ConfOutput:
                 raise(Exception(f"input_path is required for lookups, but received [{self.input_path}]"))
 
             files = glob.iglob(os.path.join(self.input_path, 'lookups', '*.csv'))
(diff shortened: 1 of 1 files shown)


pyth0n1c:
Hello @kyokkygo . Thank you for the PR and feedback.  We have actually merged this with some other changes into a different PR. You can see the respective changes at the following link:
https://github.com/splunk/contentctl/pull/34/files#diff-66d0bb652d72be60ceb4b699fa55e9c053761acad73e6b48af80e3f708d53c9a

As such, as I going to close this PR. But please be assured that these changes will make it into contentctl in the next few days as we merge the above changes.
Thank you!

PR #48, opened 2023-08-18
Allow individual detections to override actions

linuxdaemon:
This allows a detection to define response actions such as
notable, email, etc to override the global defaults.

This also adds a config option to copy the default config in to the
detection yaml on generation, to allow easier editing afterwards

### DIFF
--- a/contentctl/contentctl.py
+++ b/contentctl/contentctl.py
@@ -220,7 +220,7 @@ def doc_gen(args) -> None:
 
 
 def new_content(args) -> None:
-
+    config = start(args)
     if args.type == "detection":
         contentType = SecurityContentType.detections
     elif args.type == "story":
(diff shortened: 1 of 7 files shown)


pyth0n1c:
I am closing this out since it is now already possible to override deployments in a detection. For example, please see what we do with Baselines as a reference:
https://github.com/splunk/security_content/blob/9f85603faa6c39c8e3113d7ab7d2f73555ead92d/baselines/baseline_of_cloud_instances_destroyed.yml#L52
Note that this is also possible with detections due to how the deployment is determined:
1. Is there a `deployment` section? If so, parse it out and use the values in it.
2. If not, get the deployment based on the `type` of the search
See how that happens here: https://github.com/splunk/contentctl/blob/5abb8e0fe1ca63fbab75b1173d6b088d95528074/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L540
which leads to:
https://github.com/splunk/contentctl/blob/5abb8e0fe1ca63fbab75b1173d6b088d95528074/contentctl/objects/deployment.py#L30

PR #53, opened 2023-08-28
Handle multi-line searches and other fields

linuxdaemon:
This ensures that a multi-line search, description, or any other field,
will not generate invalid splunk config files

### DIFF
--- a/contentctl/output/conf_writer.py
+++ b/contentctl/output/conf_writer.py
@@ -54,7 +54,9 @@ class ConfWriter():
         j2_env = Environment(
             loader=FileSystemLoader(os.path.join(os.path.dirname(__file__), 'templates')), 
             trim_blocks=True,
-            undefined=StrictUndefined)
+            undefined=StrictUndefined,
+            finalize=lambda var: var if not isinstance(var, str) else var.strip().replace('\n', '\\\n')
+        )
 
 
         j2_env.filters['custom_jinja2_enrichment_filter'] = custom_jinja2_enrichment_filter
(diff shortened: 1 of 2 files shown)


pyth0n1c:
Thank you for this! Historically, accidental newlines making it into conf files has been a headache in the past for us.
We've recently added some logic that is intended to be applied to fields which contain long strings of text, like description, how_to_implement, false_positives, and search that checks for newline issues during contentctl validate.

Using this new validation, ANY newline in one of this fields NOT prefixed by a \ will result in a validation error.  We've done it this way to make it EXTRA clear, in the YML, when a newline should actually be included.  Some of our text fields wind up using different YML formatting (which can be a bit confusing), so we erred on the side of being more explicit:
https://yaml-multiline.info/


@linuxdaemon , do you have a personal preference for catch these issues during contentctl validate VS making and informed decision when the conf file is written (like you have in this PR)?

pyth0n1c:
Hi @linuxdaemon - we have implemented this change using your suggestion (usr RawConfigParser) as part of a larger change:
1. Removing all `\`- escaped YML fields in the repo.  These were a common cause of troubles because, while they helped in writing out conf files, they made it much more difficult when these YML files were rendered elsewhere such as [https://research.splunk.com](https://research.splunk.com)
2.  The logic that performs serialization to conf outputs now appends a `\t` at the end of each newline, making it more readable.  A `\` is no longer included at the end of the line in a conf file:

3.  After writing out all conf files, there are read back using the RawConfigParser as suggested. While the steps above should prevent errors in the generation of the conf file, any errors this read operation throws will make deficiencies in our output CONF files very apparent:
https://github.com/splunk/contentctl/pull/90/commits/cfe6178de1fa119842cbe3836a06ab773c2723c4 

As such I am closing this PR as the functionality will be merged when the following PR is merged:
https://github.com/splunk/contentctl/pull/90

PR #64, opened 2023-10-19
Add appinspect rest support

pyth0n1c:
Allow contentctl build and contentctl test to directly run AppInspect via the REST API.
This eliminates the need for users to write their own shell script (or otherwise) to accomplish this vetting before deploying to their own Splunk Cloud Environment.


patel-bhavin:
Tested locally and the report look fine :)

image

patel-bhavin (approved):

PR #65, opened 2023-10-20
Additional changes for BA

ljstella:
Includes changes to `contentctl/input/backend_splunk_ba.py`, some of which are present in #63 (once that merges, I will clean this up)

Primary change here is enabling the `field IN (value1, value2...)` syntax and using it for all non-wildcarded searches of the same field


pyth0n1c (approved):
Reviewed the changes in the code and reviewed generated output detections.
Looks great to me!

PR #67, opened 2023-10-20
Copy mlfiles along

pyth0n1c:
Ensure that BOTH lookups and mlmodel
files are copied to the lookups folder in
the app when it is built.


patel-bhavin (approved):

patel-bhavin:
Nice catch, PR looks good!

PR #71, opened 2023-10-25
Diff test cli support

pyth0n1c:
improved support for --mode changes testing


patel-bhavin:
Works correctly when the contentctl_test.yml is configured correctly.  and tested this PR with mode=changes

patel-bhavin (approved):

PR #76, opened 2023-11-08
Updating Apps to test detections

patel-bhavin:
This PR updates all the default apps that get loaded onto the Splunk Server for testing detections

- Uses the latest TA versions available on splunkbase
- Updated relevant Attack Range app binaries


pyth0n1c (approved):
Much better structure.  All apps appear valid and to be latest versions.

PR #77, opened 2023-11-09
detection_type = STREAMING

patel-bhavin:
An ask from BA team to introduce `detection_type = STREAMING` as a top level key to the yaml

This PR adds a detection type when the yamls are generated into dist/ssa

TR: TR-3490

Output
image


pyth0n1c (approved):
Looks good to me. If an object ALWAYS contains the same static value, I think we should have a discussion about whether that static value should be set in the Pydantic Model vs. at output generation time.
Given that other static fields are set at output generation time, this update looks good.

PR #80, opened 2023-11-15
removing file_path_key from SSE API files

patel-bhavin:
This PR removes the "file_path" key from the all jsons that are generated for SSE


ljstella (approved):
LGTM, tested w/ latest develop of security_content, reduced file size of detections.json

PR #82, opened 2023-11-22
Lookup validation - Ignore local=true

0xC0FFEEEE:
We've modified a couple of our Azure AD rules to use `local=true` as the ESCU searches fail on our cloud ES search head.

After converting our savedsearches.conf search back to YAML and running validation over the rules, contentctl complains that the `local=true` lookup doesn't exist. This simple change adds an an additional non-capture group to ignore this option.


pyth0n1c:
I think we will have to make this (and the larger lookup matching regex) a bit more robust against reordering parameters: https://github.com/splunk/contentctl/pull/82/files#diff-4ade9637ffaa744323786af0102ad18f3f46ea560f36f867aba8cc66c05d171aR51 
It will be a fairly complicated regex and will take a bit of experimentation. I'll reach out to someone on my team who is far better at regexes than I am to refine it :)

pyth0n1c:
This has turned into a much larger PR than I originally thought. We will work towards supporting things like reordering, inputlookup, and outputlook as well.
As such, I am closing this one out in favor of the new one: https://github.com/splunk/contentctl/pull/274

PR #79, opened 2023-11-13
Metadata for BA

ljstella:
Adding source and sourcetype to a finding report's evidence object.

Context: TR-3481, TR-3482

PR #85, opened 2023-11-27
[TR-3481] Sourcetype output

ljstella:
Adding the `sourceType` output in BA Finding Reports for TR-3481. This is good to go whenever.
#79  will be held until BA is ready for TR-3482.


pyth0n1c (approved):
Spoke with PR author about a minor fix which has already been made. Looks good!

PR #87, opened 2023-11-28
Small fix for finding reports 

ljstella:
quotes are hard to see


pyth0n1c (approved):
This was an easy bug to miss, thanks for the fix! Looks good!

PR #88, opened 2023-11-30
Adding integration testing to contentctl

cmcginley-splunk:
# Context
* Integration testing used to live in attack_range
* Now, it lives in contentctl :)

# Notable design decisions
Besides just adding integration testing, a few of fairly significant design changes were made to accommodate it
* Tests now have an added layer of organization called `TestGroup`
  * This unit facilitates the association of a unit test with an integration test, so that they can share the same attack data and cleanup only occurs after all tests in the `TestGroup` complete
* A notion of `status` (PASS, FAIL, SKIP, ERROR) was added to all test results
  * This gives us powerful insights when reviewing test runs, as a test which failed due to a lack of expected results is very different to investigate compared to a test which encountered an unhandled exception
  * This also puts our test results in line w/ the use of status by other tools which ingest our results for engineers to review
  * Also useful for identifying when an integration test was skipped (e.g. integration testing was fully suppressed, or an individual integration test was skipped due to the detection being of type  Hunting)
  * `success` is now a property derived from status
    * If `status` is PASS or SKIP, success is True
    * If `status` is ERROR or FAIL, success if False
   * An overall detection is considered successful if it had tests and all of its tests were successful (all states PASS or SKIP)
   * An overall test run is considered successful if all detections were successful
* The count for untested and failed detections were separated into two fields in the results summary
* In the report YAML, each tests shows its duration in terms of how long the test itself took PLUS the setup/cleanup. So, in this context, setup/cleanup time is "double-counted" across tests in the same group. I'm open to reporting this in a different way, this was just the decision I made based on the newly shared setup/cleanup time across tests

# Code changes
* Integration testing added (backed primarily by the `CorrelationSearch` class
* Integration tests and `TestGroups` are created implicitly when a `Detection` is built, based on each unit test (one integration test and one `TestGroup` per unit test)
* Integration testing is disabled by default, enabled at either the CLI or in the config with `enable_integration_testing`
* Test execution now iterates over `TestGroup`s instead of individual tests
* All test results now have a status (PASS, FAIL, SKIP, ERROR)
* Some of the pbar code was pulled into a utility module called `progress_bar`; this was so that `CorrelationSearch` could have access to the same logging outside of `DetectionTestingInfrastructure`
* Pbar now reports testing w/ an additional indicator (SETUP, UNIT, INTEG, or GROUP)
* Consolidated reporting states under enums
* To facilitate testing organized in groups, the reporting of test duration had to be changed a bit
  * At the CLI, each test show only the duration for how long the test itself took, excluding data replay and cleanup. Group reporting at the CLI shows how long all tests took, plus the setup/cleanup
* Many small quality improvements around typing and formatting based on reports from flake8 and mypy

# Future work
* Tracked and shared elsewhere on a ticket internally


pyth0n1c (approved):
Lots of big changes, awesome work to @cmcginley-splunk .
I have reviewed everything and run extended local testing with and without ES/ --enable_integration_testing flag.
Approved with no changes.
I have bumped the version in pyproject.toml and will release this as version 3.1.0 to pypi

PR #90, opened 2023-12-07
Pydantic2

pyth0n1c:
Move to Pydantic2 and massively increase typing, inheritance/code reuse, and inline documentation for types around all contentctl objects.

Reduce the number of custom validators and leverage more built-in types, Enumerations, and Pydantic Field validators (for example, using regular expressions). 

This will also move to Pydantic2, allowing significantly better, faster validation.  In many cases, speedups of 5-50x are possible!


ljstella (approved):
LGTM-

Things to cover before a tagged release:
1) Once over of README.md, specifically examples
2) confirm exactly which functionality we've at least temporarily removed (which should likely cause a major version bump)

Things to figure out later:
1) Determining if we're going to add back any of the functionality we've removed

:shipit:

contentctl/objects/config.py
```diff
@@ -1,110 +1,106 @@
...
+    def ensureAppPathExists(self, config:test, stage_file:bool=False):
+        if stage_file:
+            if not config.getLocalAppDir().exists():
+                config.getLocalAppDir().mkdir(parents=True)
 
-class ConfigBuildSplunk(BaseModel):
-    pass
+class TestApp(App_Base):
+    model_config = ConfigDict(use_enum_values=True,validate_default=True, arbitrary_types_allowed=True)
+    hardcoded_path: Optional[Union[FilePath,HttpUrl]] = Field(default=None, description="This may be a relative or absolute link to a file OR an HTTP URL linking to your app.")
```
cmcginley-splunk, line 55: What's the use case for making `hardcodedpath` `Optional`? In what circumstances would you want to define an app w/o a URL or fielpath?

contentctl/objects/config.py
```diff
@@ -1,110 +1,106 @@
...
-    pass
-
-class ConfigBuildBa(BaseModel):
-    pass
-
 
-
-class ConfigBuild(BaseModel):
+    @field_serializer('hardcoded_path',when_used='always')
+    def serialize_path(path: Union[AnyUrl, pathlib.Path])->str:
```
cmcginley-splunk, line 59: Why is the typing `Union[AnyUrl, pathlib.Path]` here but `Optional[Union[FilePath,HttpUrl]]` in the attribute definition?

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructureContainer.py
```diff
@@ -119,7 +133,7 @@ def removeContainer(self, removeVolumes: bool = True, forceRemove: bool = True):
         try:
             # If the user wants to persist the container (or use a previously configured container), then DO NOT remove it.
             # Emit the following message, which they will see on initial setup and teardown at the end of the test. 
-            if self.global_config.infrastructure_config.persist_and_reuse_container:
+            if self.global_config.container_settings.leave_running:
                 print(f"\nContainer [{self.get_name()}] has NOT been terminated because 'contentctl_test.yml ---> infrastructure_config ---> persist_and_reuse_container = True'")
```
cmcginley-splunk, line 137: This line references a portion of the config which no longer exists

contentctl/helper/config_handler.py
cmcginley-splunk, line 1: Are we planning on retaining the `read_test_config` function for now?

PR #92, opened 2024-01-04
Adding risk message validation++

cmcginley-splunk:
# Context
* As part of PEX-363, we wanted to expand integration testing to validate risk message content
* If invalid fields are referenced in the risk message, they will fail to render silently within ES

# Code changes
* Added new `RiskEvent` model for the events returned by ES
* Refined search criteria in`CorrelationSearch` to only search for events which match the appropriate search name
* Risk validation expanded to check:
  * Risk message contains no `$...$` literals (**now scans for the specific field replacement tokens rather than the `$...$` pattern)
  * Risk message in detection matches risk message in risk event (regex conversion)
  * Risk score matches detection
  * Analytic stories match detection
  * MITRE ATT&CK IDs match detection
* Added new `NotableEvent` model for the events returned by ES (will support additional notable validation, potentially added in the future)
* Removed a bug where we we were sleeping for 60s initially twice (should remove 20h from our cumulative compute time, about 30min on our total runtime across 40 instances)
* Refined cleanup logic
* Refactored `format_pbar_string` s.t. it uses the `start_time` instance attribute if none is provided explicitly
* Linted/formatted/typed:
  * `observable.py`
  * `detection_abstract.py`
  * `security_content_object_abstract.py`
  * `atomic.py`
  * `base_test_result.py`
  * `detection_tags.py`
* Added debug configuration where you can select your args; must supply path to `security_content` repo as an env var (SECURITY_CONTENT_PATH)

# New detection failures
~~I spot checked these, but did not do a deep dive on every single one; but I believe they are all legitimate validation issues~~
* ~~**59 detections** are failing this new validation (tested on v4.29.0)~~
  * ~~51 detections create risk events where the risk message contains a `$...$` literal -> represents a bad field substitution, likely because the referenced field doesn't exist in final SPL output~~
  * ~~8 detections have mismatches between the analytic stories listed in the detection and what's observed in the risk event -> spot checking shows this is mostly due to typos/casing in analytic story names in the detection which fails to link the actual analytic story to the risk event in ES~~
 
~~**NOTE**: this testing was performed locally, and some detections failed due to networking issues, likely due to my ISP bandwidth; so there may be more legitimate failures similar to the above not captured in this initial test~~

* **UPDATE 8/7/24**: 
  * Ran full package test on develop
  * Three detections failed risk message validation after refinement (tracked for resolution internally)
  * All false positives have been eliminated w/ the new method of scanning the risk_message for the field replacement tokens instead of the `$...$` pattern

# Testing
Will post some results from an SCA pipeline when that run completes

# Future work
* This PR also adds some commented out code which matches risk events to observables in the detection
  * This feature is still in progress as it does generate some false positives
  * That said, it did expose some legitimate issues (e.g. Attacker role observables creating risk events), so it will be a good addition once that validation logic has been refined to remove false positives.
* **Add validation which enforces the usage of field names w/o spaces; until this is enforced, risk message validation will not be able to validate risk messages leveraging such fields as the extraction pattern we use for pulling the field tokens to scan for assumes no spaces are present**
* **Add validation which checks for balanced `$` in `tags.message`; if a need for a `$` literal in messages is found, we can consider adding some kind of escaping to support it**


contentctl/objects/correlation_search.py
```diff
@@ -18,10 +19,22 @@
...
+from contentctl.objects.risk_event import RiskEvent
+from contentctl.objects.notable_event import NotableEvent
+from contentctl.objects.observable import Observable
+
+# TODO: test my new contentctl logic against an old ESCU build; my logic should detect the faulty attacker events
 
 
 # Suppress logging by default; enable for local testing
-ENABLE_LOGGING = False
+ENABLE_LOGGING = True
```
cmcginley-splunk, line 37: Will set this back to False after a final full test w/in an SCA pipeline

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -416,7 +412,7 @@ def test_detection(self, detection: Detection) -> None:
         """
         # TODO: do we want to return a failure here if no test exists for a production detection?
```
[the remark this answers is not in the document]
patel-bhavin: Yes, contentctl validate fails today, when a production detection does not have tests: Verified by removing tests from production detection
image

contentctl/objects/correlation_search.py
patel-bhavin: @cmcginley-splunk : Should we set this back to false now that we have ran this a few times in SCA repo?

cmcginley-splunk:
Some detections seem to be generating false positives (test failures that are not actual failures); e.g.: Windows Excessive Disabled Services Event

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -236,6 +234,8 @@ def connect_to_api(self, sleep_seconds: int = 5):
                 self.pbar.write(
                     f"Error getting API connection (not quitting) '{type(e).__name__}': {str(e)}"
                 )
+                # TODO (cmcginley): can we remove this print? Can we comment on this commented out
+                #   code?
                 print("wow")
```
cmcginley-splunk, line 239: Can this be removed?

contentctl/enrichments/attack_enrichment.py
```diff
@@ -26,6 +26,10 @@ def getAttackEnrichment(config:validate)->AttackEnrichment:
         return enrichment
     
     def getEnrichmentByMitreID(self, mitre_id:Annotated[str, Field(pattern="^T\d{4}(.\d{3})?$")])->Union[MitreAttackEnrichment,None]:
+        # TODO (cmcginley): do we want to return None when use_enrichment is false? This seems
+        #   confusing, and it seems that we intentionally do not call this function when that
+        #   setting is enabled; I'd vote to have this raise if this flag is set and this function
+        #   is called
         if not self.use_enrichment:
```
cmcginley-splunk, line 33: do we want to return None when use_enrichment is false? This seems confusing, and it seems that we intentionally do not call this function when that setting is enabled in `DetectionTags.addAttackEnrichment`. I'm unclear why we have a separate invocation of this in `SSADetectionBuilder`. I'd vote to streamline the usage of this and have this raise if this flag is set and this function is called.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -123,98 +141,97 @@ def validate_test_groups(cls, value:Union[None, List[TestGroup]], info:Validatio
...
             return []
-    
 
     @computed_field
     @property
-    def source(self)->str:
+    def source(self) -> str:
+        # TODO (cmcginley): if filepath cannot be None for a detection, we should be enforcing that
+        #   with a validator and a default value/factory
         if self.file_path is not None:
```
cmcginley-splunk, line 168: We have `file_path` as a required field; this check is irrelevant no? Pydantic validation precludesthe need for this check

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -223,42 +240,40 @@ def cve_enrichment_func(self, info:ValidationInfo):
                 raise ValueError(f"{e}")
         self.cve_enrichment = enriched_cves
         return self
-    
 
+    # TODO (cmcginley): can we tighten this typing? is this field even being used?
     splunk_app_enrichment: Optional[List[dict]] = None
```
cmcginley-splunk, line 245: Is this field even used?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -338,171 +356,209 @@ def serialize_model(self):
...
 
+        # return the model
+        return super_fields
 
-    def model_post_init(self, ctx:dict[str,Any]):
-        # director: Optional[DirectorOutputDto] = ctx.get("output_dto",None)
+    def model_post_init(self, __context: Any) -> None:
+        super().model_post_init(__context)
+        # TODO (cmcginley): can we remove this null/type check back?
+        # director: Optional[DirectorOutputDto] = __context.get("output_dto",None)
```
cmcginley-splunk, line 397: can we remove this commented out null type check?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -338,171 +356,209 @@ def serialize_model(self):
...
+            )
+
         self.data_source_objects = matched_data_sources
 
         for story in self.tags.analytic_story:
-            story.detections.append(self)            
-        return self
+            # TODO (cmcginley): I think this instance of typing weirdness highlights some of the
+            #   issues I see with maintaining Detection and Detection_Abstract
+            story.detections.append(self)
```
cmcginley-splunk, line 454: This is an instance of some of the weird typing situations we run into w/ `Detection` vs `Detection_Abstract`. `self` is of type `Detection_Abstract` but `story.detections` is declared as a list of `Detection` objects. This part of the `model_post_init` should be moved to the `Detection` class; or, better yet, we should remove the `Detection_Abstract` class altogether IMO.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -338,171 +356,209 @@ def serialize_model(self):
...
-            raise ValueError("Error, baselines are constructed automatically at runtime.  Please do not include this field.")
+            raise ValueError(
+                "Error, baselines are constructed automatically at runtime.  Please do not include this field."
+            )
 
-        
-        name:Union[str,None] = info.data.get("name",None)
+        # TODO (cmcginley): if it's possible in practice for this to return None, we should adjust
+        #   the declared type
+        name: Union[str, dict] = info.data.get("name", None)                                        # type: ignore
```
cmcginley-splunk, line 487: If it's possible in practice for this to return `None` that should be part of the type, otherwise the check on line 488 makes no sense.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -338,171 +356,209 @@ def serialize_model(self):
...
-    
+
     @field_validator("deployment", mode="before")
-    def getDeployment(cls, v:Any, info:ValidationInfo)->Deployment:
-        return Deployment.getDeployment(v,info)
-        return SecurityContentObject.getDeploymentFromType(info.data.get("type",None), info)
-        # director: Optional[DirectorOutputDto] = info.context.get("output_dto",None) 
+    def getDeployment(cls, v: Any, info: ValidationInfo) -> Deployment:
+        return Deployment.getDeployment(v, info)
+        return SecurityContentObject.getDeploymentFromType(info.data.get("type", None), info)
```
cmcginley-splunk, line 556: can we remove any of this dead code?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -338,171 +356,209 @@ def serialize_model(self):
...
-        if not isinstance(search,str):
-            #The search was sigma formatted (or failed other validation and was None), so we will not validate macros in it
+    def getDetectionMacros(cls, v: list[str], info: ValidationInfo) -> list[Macro]:
+        if info.context is None:
+            raise ValueError("ValidationInfo.context unexpectedly null")
+
+        director: DirectorOutputDto = info.context.get("output_dto", None)
+
+        # TODO (cmcginley): what's the typing on this dict?
+        search: Union[str, dict] = info.data.get("search", None)
```
cmcginley-splunk, line 513: if search is a dict, what's the possible typing? is it `dict[str, str]`?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -516,50 +572,56 @@ def getDeployment(cls, v:Any, info:ValidationInfo)->Deployment:
...
+        detection_filenames: set[str],
+        all_detections: list[Detection_Abstract]
+    ) -> list[Detection_Abstract]:
         detection_filenames = set(str(pathlib.Path(filename).absolute()) for filename in detection_filenames)
-        detection_dict = SecurityContentObject.create_filename_to_content_dict(all_detections)
+        detection_dict = SecurityContentObject.create_filename_to_content_dict(all_detections)      # type: ignore
 
         try:
+            # TODO (cmcginley): incompatible return type
             return [detection_dict[detection_filename] for detection_filename in detection_filenames]
```
cmcginley-splunk, line 585: We are returning a list of `SecurityContentObject_Abstract`; we should be returning `Detection_Abstract`; perhaps `Detection_Abstract` should have an overwridden version of `create_filename_to_content_dict` to accomodate this?

contentctl/objects/abstract_security_content_objects/security_content_object_abstract.py
```diff
@@ -1,31 +1,39 @@
...
-from typing import TYPE_CHECKING, Self
+from typing import TYPE_CHECKING, Self, Any
 
+# TODO (cmcginley): this config object does not seem to exist and is not used anywhere in this
+#   file; can we remove?
 if TYPE_CHECKING:
-    from contentctl.objects.deployment import Deployment        
+    from contentctl.objects.deployment import Deployment
     from contentctl.objects.security_content_object import SecurityContentObject
     from contentctl.objects.config import Config
```
cmcginley-splunk, line 9: Can we remove this `Config`? It's not used anywhere and does not seem to exist.

Additionally, what's happening here w/ the check against `TYPE_CHECKING`?

contentctl/objects/detection_tags.py
```diff
@@ -1,132 +1,151 @@
...
         if self.security_domain == SecurityDomain.NETWORK:
             return [Cis18Value.CIS_13]
         else:
             return [Cis18Value.CIS_10]
 
-    
     research_site_url: Optional[HttpUrl] = None
     event_schema: str = "ocsf"
+    # TODO (cmcginley): mappings should be fleshed out into a proper class
     mappings: Optional[List] = None
```
cmcginley-splunk, line 110: I believe this should be a list of of `dict[str, str]` yes? Perhaps we should make a model for `Mapping` since it has consistent key structure?

contentctl/objects/lookup.py
```diff
@@ -25,6 +25,7 @@
 class Lookup(SecurityContentObject):
     
     collection: Optional[str] = None
+    # TODO (cmcginley): should fields_list be a CSV string? or a list of strings?
     fields_list: Optional[str] = None
```
cmcginley-splunk, line 29: We do we parse `fields_list` as a CSV string and not a list of strings?

contentctl/objects/unit_test.py
```diff
@@ -1,14 +1,14 @@
...
 if TYPE_CHECKING:
     from contentctl.objects.unit_test_attack_data import UnitTestAttackData
     from contentctl.objects.unit_test_result import UnitTestResult
 
-from typing import Union
-
 from pydantic import Field
 
+# TODO (cmcginley): can these be removed?
 # from contentctl.objects.security_content_object import SecurityContentObject
```
cmcginley-splunk, line 12: Do we need this commented out code?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: Agreed, I think was a relic from a time when file_path was not a required_field (and that functionality doesn't even work right now, since it raises an exception).  I have removed that check and now just returned the parent name.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: Yup, that's correct! I will make this change as well.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: Yes, I have removed all that dead code.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: The searches that are dicts are sigma searches, so the typing for those should be dict[str,Any]. I will update that typing, and the typing for the search field, to reflect that.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: This function is actually not used anymore, so it has been removed.

contentctl/objects/abstract_security_content_objects/security_content_object_abstract.py
pyth0n1c: Yes, this has now been removed.

contentctl/objects/baseline_tags.py
```diff
@@ -17,6 +17,7 @@
 class BaselineTags(BaseModel):
     analytic_story: list[Story] = Field(...)
     #deployment: Deployment = Field('SET_IN_GET_DEPLOYMENT_FUNCTION')
+    # TODO (cmcginley): can we remove str from the possible types here?
     detections: List[Union[Detection,str]] = Field(...)
```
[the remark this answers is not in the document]
pyth0n1c: This typing is weird, but intentional.  In the baseline files, we define the related searches initially as a list of strings.  Then, we construct the Baseline objects.  However, at this time, the related Detection objects do not exist!

Then, later, we construct the detections and iterate over the Baselines and convert the list of detection name strings into a list of Detection OBJECTS. You can see the code for this here:
https://github.com/splunk/contentctl/blob/96b813fa220ffef6349487421687b99404f647a3/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L401-L415

You will notice that because Detections are constructed after Baselines, Macros, Lookups, etc, they don't have this weird double-typing:  https://github.com/splunk/contentctl/blob/96b813fa220ffef6349487421687b99404f647a3/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L199

I would love to do this a different way, but can't think of it - I am very open to suggestions.

contentctl/objects/detection_tags.py
pyth0n1c: ABSOLUTELY. Anything we can do to move away from `dicts` is a huge win in my book.
I will leave this one open as I haven't made that change yet.

contentctl/objects/detection_tags.py
```diff
@@ -141,112 +160,125 @@ def serialize_model(self):
...
             raise ValueError("Context not provided to detection.detection_tags.atomic_guid validator")
 
-        
-        all_tests:List[AtomicTest]= output_dto.atomic_tests
-        
-        matched_tests:List[AtomicTest] = []
-        missing_tests:List[UUID4] = []
-        badly_formatted_guids:List[str] = []
+        # TODO (cmcginley): typing issue; what to do if atomic_tests is None?
+        all_tests: List[AtomicTest] = output_dto.atomic_tests
```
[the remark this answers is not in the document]
pyth0n1c: I have changed the typing for DirectorOutputDto.atomic_tests to be list[AtomicTest] instead. Now, atomic_tests will be an empty list if we have not read any from the repo. I have verified that this change still works as expected whether enrichment is enabled/disabled.

contentctl/objects/lookup.py
pyth0n1c: This is probably somewhat-legacy code/typing.  In the YMLs, fields is a comma-separated string. We should likley convert all the YMLs to use a list[str].
Additionally, Lookups are actually two different kinds of lookups - CSVLookup and KVStore Lookup. These have have different requirements. When we make the above change, we should consider breaking this into two different classes with Lookup as an Abstract Base Class.

I will leave this comment open for discussion

contentctl/objects/unit_test.py
pyth0n1c: Yes, they have been removed.

contentctl/objects/lookup.py
cmcginley-splunk: How do we want to track small quality of life improvements like this? GitHub issues? Internally on Jira?

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
cmcginley-splunk: Removed this where? I don't see the change on this branch

contentctl/objects/lookup.py
cmcginley-splunk: See https://github.com/splunk/contentctl/issues/220

contentctl/objects/detection_tags.py
cmcginley-splunk: See https://github.com/splunk/contentctl/issues/221

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -236,6 +234,8 @@ def connect_to_api(self, sleep_seconds: int = 5):
...
                 # self.pbar.write(
```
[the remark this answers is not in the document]
pyth0n1c: Yes, this has been removed.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: This is legacy code which is related to DocGen Functionality, which is no longer exposed. I have raised an internal issue with the team to see if DocGen can/should be removed. If so, that will be a much larger PR but should be a nice cleanup to numerous files :) 
As such, I am going to leave this comment unresolved, but no relevant changes should be part of THIS PR.

contentctl/objects/unit_test.py
```diff
@@ -1,16 +1,13 @@
 from __future__ import annotations
-from pydantic import Field
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
+
+# TODO (cmcginley): what's the purpose of this conditional?
```
cmcginley-splunk, line 4: What're we doing here w/ these conditional imports? Esp. bc we import them again a few lines later

contentctl/objects/baseline_tags.py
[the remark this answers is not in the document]
cmcginley-splunk: Understood! Will investigate later as part of #223

contentctl/objects/abstract_security_content_objects/detection_abstract.py
cmcginley-splunk: Resolving for now; I think theres a larger design convo to be had about `Detection` and `Detection_Abstract`

contentctl/objects/unit_test.py
pyth0n1c: Good catch. These should not be double-imported.
The conditional imports are because many of our objects are structured...interesting and _intentionally_ do circular imports.  For example, a Story includes a list of Detections and a Detection includes a list of Stories.

For typing reasons, Story needs to import Detection and Detection needs to import Story - but Python/Pydantic throw parse exceptions if you don't import them (as expected). However, a regular import throws a _different_ runtime Pydantic exception about models not being fully defined.

Putting them imports in an if `TYPE_CHECKING` block satisfies the type checker and avoids the runtime errors mentioned above. It's a bit of a mess - we may be able to do this better by _explicitly_ calling `PydanticObject.model_rebuild` after all objects have been imported, but I haven't been able to resolve that yet.

pyth0n1c (approved):
Casey and I have discussed this and made some improvements based on the suggested feedback. We have also opened some issues for ongoing, related work.
This is approved - we will wait on the outcome of some internal testing before actually merging it in.

pyth0n1c (approved):
I bumped the version, which dismissed my previous approval. Just approving again

patel-bhavin (approved):

PR #94, opened 2024-01-19
Adding Release Notes Function 🛺 

patel-bhavin:
This MR adds code to compare a git tagged branch with the previous one to generate release notes specific for ESCU app 
🎶 

Usage: contentctl release_notes --tag v4.20.0

image


0xC0FFEEEE:
Neat feature!

Couple of observations from an interested party who has spent too much time perusing this codebase...

- `repo_path` is currently hard-coded in a couple of places, this should probably use args.path supplied via the -p/--path flags.
- The features logic, `create_notes` and `release_note` should also be spun in to a submodule within the `actions/` directory.

patel-bhavin:
@0xC0FFEEEE : Dan, thanks for taking a look and providing feedback! Agreed on both points, i will work with @pyth0n1c to get those points addressed.

PR #97, opened 2024-01-31
Fix manual test bug

pyth0n1c:
Respect manual_testing flag.
Previously, there was a bug where the validator was not always working as expected because:
1.  If tests were not defined, then the validator was not called. This has been fixed by setting it to always=True
2.  Because of the way tests are run and results are generated, it was showing as a failure. We now show a manual test as a "successful" test with a special message indicating that it was marked as "manual_test" in the output in test_results/summary.yml. Please see the attached image as an example. This should make it easy to review results.

Note that test 'name' and 'message' fields have been dynamically generated at run time due to the presence of the manual_test field:
image

### DIFF
--- a/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
+++ b/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
@@ -375,7 +375,20 @@ class DetectionTestingInfrastructure(BaseModel, abc.ABC):
                 self.sync_obj.currentTestingQueue[self.get_name()] = None
 
     def test_detection(self, detection: Detection):
-        if detection.tests is None:
+        if detection.tags.manual_test:
+            if len(detection.tests) != 1:
+                raise Exception(f"Detection {detection.name} configured for manual_test, but no "
+                                "manual_test was found in tests section. It should have been "
+                                f"added automatically: {detection.tests}")
+            elif detection.tests[0].name != "MANUAL_TEST_ONLY":
+                raise Exception(f"Detection {detection.name} configured for manual_test, but a test "
+                                f"{detection.tests[0]}was found INSTEAD of the automatically added "
+                                "MANUAL_TEST_ONLY")
+            else:
+                detection.tests[0].result = UnitTestResult()
+                detection.tests[0].result.set_manual_test(detection.tags.manual_test)
+                return
+        elif detection.tests is None:
             self.pbar.write(f"No test(s) found for {detection.name}")
             return
 
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -161,12 +161,26 @@ class Detection_Abstract(SecurityContentObject):
         # Found everything
         return v
 
-    @validator("tests")
+    @validator("tests", always=True)
     def tests_validate(cls, v, values):
-        if values.get("status","") == DetectionStatus.production.value and not v:
-            raise ValueError(
-                "At least one test is REQUIRED for production detection: " + values["name"]
-            )
+        #Only production analytics require tests
+        if values.get("status","") != DetectionStatus.production.value:
+            return v
+        
+        #Only certain types of analytics must have tests
+        if values.get("type","") not in set([AnalyticsType.Anomaly.value, AnalyticsType.Hunting.value, AnalyticsType.TTP.value]):
+            return v
+        
+        #check and see if the manual_test flag is present
+        tags_obj = values.get("tags")
+        if tags_obj and tags_obj.manual_test:
+            #If any tests are present while manual_test is set, replace them with the manual_test placeholder
+            return [{"name":"MANUAL_TEST_ONLY"}]
+
+        if len(v) < 1:
+            raise ValueError("At least one test is REQUIRED for production detection: " + values.get("name", "NO NAME FOUND"))
+        
+        #No issues - at least one test provided            
         return v
     
     @validator("datamodel")
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -175,7 +175,7 @@ class Detection_Abstract(SecurityContentObject):
         tags_obj = values.get("tags")
         if tags_obj and tags_obj.manual_test:
             #If any tests are present while manual_test is set, replace them with the manual_test placeholder
-            return [{"name":"MANUAL_TEST_ONLY"}]
+            return [UnitTest(name="MANUAL_TEST_ONLY",attack_data=[])]
 
         if len(v) < 1:
             raise ValueError("At least one test is REQUIRED for production detection: " + values.get("name", "NO NAME FOUND"))
--- a/contentctl/objects/unit_test_result.py
+++ b/contentctl/objects/unit_test_result.py
@@ -54,6 +54,15 @@ class UnitTestResult(BaseModel):
 
         return results_dict
 
+    def set_manual_test(self,manual_test_text:str):
+        self.duration = 0
+        self.exception = None
+        self.message = f"manual_test: {manual_test_text}"
+        self.success = True
+        self.job_content = None
+        self.sid_link = "manual_test - NO Search ID"
+        return self.success
+    
     def set_job_content(
         self,
         content: Union[Record, None],


contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -375,7 +375,20 @@ def execute(self):
                 self.sync_obj.currentTestingQueue[self.get_name()] = None
 
     def test_detection(self, detection: Detection):
-        if detection.tests is None:
+        if detection.tags.manual_test:
+            if len(detection.tests) != 1:
```
cmcginley-splunk, line 379: The edge case of there being 2 or more tests that were added, which I think you'd still want to validate against

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -161,12 +161,26 @@ def search_obsersables_exist_validate(cls, v, values):
...
-        if values.get("status","") == DetectionStatus.production.value and not v:
-            raise ValueError(
-                "At least one test is REQUIRED for production detection: " + values["name"]
-            )
+        #Only production analytics require tests
+        if values.get("status","") != DetectionStatus.production.value:
+            return v
+        
+        #Only certain types of analytics must have tests
+        if values.get("type","") not in set([AnalyticsType.Anomaly.value, AnalyticsType.Hunting.value, AnalyticsType.TTP.value]):
```
cmcginley-splunk, line 171: nit: I'd personally rather see this as a check against the list of the types which don't need tests, rather then the analytics which DO need tests. The subtle difference would be that any new types added would require tests by default and need to be added to this list as an exception

contentctl/objects/unit_test_result.py
```diff
@@ -54,6 +54,15 @@ def get_summary_dict(
 
         return results_dict
 
+    def set_manual_test(self,manual_test_text:str):
```
cmcginley-splunk, line 57: I think marking this as a success is OK for now. But, my integration testing changes introduce a notion of states (PASS, FAIL, ERROR, SKIP) and I think SKIP would be more appropriate for the manual_tests ultimately

cmcginley-splunk (commented):
Have a few comments for you Eric, but looks good!

pyth0n1c:
Merged latest contentctl updates into this branch. Now that a test SKIPPED state has been introduced, detections marked with MANUAL_TEST will instead be SKIPPED instead of simply logged as successful

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
pyth0n1c: I believe this this has been addressed. Now, each test under a detection tagged manual_test gets status `skip` and a message describing WHY it was skipped (the text contained in the manual_test field) AND if a detection tagged `manual_test` does not have ANY tests, it gets a descriptive, fake one added at runtime showing that it was a manual_test detection.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: I agree, the code has been updated to reflect this.

contentctl/objects/unit_test_result.py
pyth0n1c: The code has now been updated to use the SKIP state. This carries through to Integration Testing as well!

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
cmcginley-splunk: Sweet! I think it's much cleaner to do the skipping elsewhere; like this change a lot

contentctl/actions/detection_testing/views/DetectionTestingView.py
```diff
@@ -93,16 +94,27 @@ def getSummaryObject(
...
-                total_pass += 1
-            else:
+            if summary["success"] is False:
                 total_fail += 1
+            else:
+                #Test is marked as a success, but we need to determine if there were skipped unit tests
+                #SKIPPED tests still show a success in this field, but we want to count them differently
+                pass_increment = 1
+                for test in summary.get("tests"):
+                    if test.get("test_type") == "unit" and test.get("status") == "skip":
```
cmcginley-splunk, line 104: I think, and I don't need this to be blocking, but I'd like to make detection state resolution (e.g. is it considered a skip/fail/error/pass) a property of the detection itself

E.g. define a property function `status` which iterates over tests
- Pass: If at least one test passes and the rest pass/skip
- Skip: If all tests skip
- Fail: If at least one test fails and the rest pass/skip/fail
- Error: If at least one detection errors (regardless of other test states)

contentctl/actions/detection_testing/views/DetectionTestingView.py
```diff
@@ -93,16 +94,27 @@ def getSummaryObject(
...
+                        break
+                total_pass += pass_increment
+                
 
             # Append to our list
             tested_detections.append(summary)
 
         # Sort s.t. all failures appear first (then by name)
-        tested_detections.sort(key=lambda x: (x["success"], x["name"]))
+        #Second short condition is a hack to get detections with unit skipped tests to appear above pass tests
```
cmcginley-splunk, line 116: Can I ask why you want skip above pass?

contentctl/actions/detection_testing/views/DetectionTestingView.py
```diff
@@ -128,14 +140,15 @@ def getSummaryObject(
...
+        total_detections = total_fail + total_pass + total_untested + total_skipped
+
 
         # Compute the percentage of completion for testing, as well as the success rate
         percent_complete = Utils.getPercent(
             len(tested_detections), len(untested_detections), 1
         )
         success_rate = Utils.getPercent(
-            total_pass, total_detections, 1
+            total_pass, total_detections-total_skipped, 1
```
cmcginley-splunk, line 151: This is a non-blocking comment and a reiteration of a conversation in slack; just including it for good book-keeping

I think future work could break down another metric as `tested_detections` which is the value you are calculating in-line here. It would provide clarity to someone reading the YAML as to how success rate is derived. Likewise, `total_detection` should become a count of all detections w/in the scope of this test run (e.g. all, selected, changes) and if all, should include experimental and deprecated in its count, with separate counts for `total_production` `total_experimental` etc.

I think this is just good data to have, but this also serves a request from Javier around having a metics for test coverage of all detections delivered in ESCU (which would include experimental and deprecated)

contentctl/input/director.py
```diff
@@ -221,6 +221,13 @@ def constructDetection(self, builder: DetectionBuilder, file_path: str) -> None:
         # TODO: is there a better way to handle this? The `test` portion of the config is not defined for validate
         if (self.input_dto.config.test is not None) and (not self.input_dto.config.test.enable_integration_testing):
             builder.skipIntegrationTests()
+        
+        if builder.security_content_obj is not None and \
```
cmcginley-splunk, line 225: This is less about this MR and more so about the class hierarchy of SecurityContentObject

Why does the DetectionBuilder class have an attribute for the SecurityContentObject and not just for a Detection?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -60,6 +60,34 @@ class Detection_Abstract(SecurityContentObject):
     class Config:
         use_enum_values = True
 
+
+    # A list of groups of tests, relying on the same data
+    test_groups: Union[list[TestGroup], None] = None
```
cmcginley-splunk, line 65: I see you migrated this code from Detection

Can you explain the intention behind having separate classes here (detection and the abstract)? I'm fine w/ this code living in either class, but I chose Detection bc I thought the spirit of the abstract class was to be a more faithful representation of the model as it exists in YAML, and since the `test_groups` attribute is an inferred field, I thought maybe it made sense to live elsewhere.

So tl;dr I'm totally fine with this, and no change required, just curious, and wondering if we don't have a need for two separate classes, perhaps we could collapse them into one in the future

contentctl/objects/unit_test_result.py
```diff
@@ -37,18 +37,18 @@ def set_job_content(
...
         if content is not None:
-            self.job_content = content
-
             if self.status == TestResultStatus.PASS:
                 self.message = "TEST PASSED"
             elif self.status == TestResultStatus.FAIL:
                 self.message = "TEST FAILED"
             elif self.status == TestResultStatus.ERROR:
-                self.message == "TEST FAILED (ERROR)"
+                self.message = "TEST FAILED (ERROR)"
```
cmcginley-splunk, line 49: nit: I personally would just list this as `TEST ERROR` for conciseness

cmcginley-splunk (dismissed):
Left some comments regarding philosophy and future work, but LGTM :)

contentctl/actions/detection_testing/views/DetectionTestingView.py
pyth0n1c: The results file can be, and often is, REALLY large.  I think it makes sense for errors/failures at the top, as these are the ones that deserve the most attention (and what people are likely to be looking for when they open this file).

I _personally_ think "pass" is the _least_ interesting since it means they are working as expected.
I think SKIP falls somewhere in between a failure and a pass.  For example, someone might want to manually evaluable these OR note how may tests have been skipped.

Definitely a personal opinion.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: Great question!  Many contentctl users want to change content somehow (and detections are the content they are most-often interested in changing).  These changes could be, but are not limited to:
1.  Adding new fields (and possibly building validators for those fields)
2.  Setting more or less strict validations on fields
3.  Setting values for fields already defined in that object (but which may be missing from their YMLs)
4. Disabling some of the validations we do.

However, if they change Detection_Abstract and WE change Detection_Abstract, it has a high probability of creating painful merge conflict issues.  As such, the plan was to retain control of Detection_Abstract and ship all our out-of-the-box functionality in that file.  Then if customers want their own functionality, they can add it to Detection.py.
This way, pulling the latest contentctl updates should be painless and without merge conflicts.  We are hoping to do this with all types, but have started with Detection since it is the one that receives user changes most often.

contentctl/objects/unit_test_result.py
pyth0n1c: Agreed, especially since that aligns with the TestResultStatus.ERROR.
I have made that change

contentctl/input/director.py
pyth0n1c: 💯 
There are huge changes/simplifications coming to the Builders in the following branch:
https://github.com/splunk/contentctl/tree/pydantic2
I am resolving this conversation because we will begin to address them there!

contentctl/actions/detection_testing/views/DetectionTestingView.py
pyth0n1c: 💯 
I think we should make this a pydantic computed_property.  On top of that, I would add an additional option to this set of values (which is the default value) along the lines of NOT_YET_TESTED.
This way, when we are rendering the current state of all tests, like in a user interface, we can easily differentiate /sort all sorts of different tests.

contentctl/actions/detection_testing/views/DetectionTestingView.py
pyth0n1c: Yes, I think this type of additional breakdown is a good idea.
I also think it should be abstracted from this particular view to a Model so that it can be used in different UI components, such as:
1. The file, like we use it today
2. Command line output
3.  Other User Interfaces we may support in the future.

PR #99, opened 2024-01-31
Create threat objects

patel-bhavin:
This MR adds extra specificity to how threat objects type are created in ss.conf 

threat object type = process, file_name, url


pyth0n1c (approved):
Reviewed and had a chat with the dev to clear up questions.
Approved!

PR #100, opened 2024-02-01
adding providing technologies field

patel-bhavin:
This PR adds providing technologies for the most of the detections that we have for okta, kubernetes, azure, etc

image


pyth0n1c (approved):
Small tweaks made to the strings searched for in the .search field.
Consulted with original developer who also approved!

pyth0n1c (approved):
Second review has no changes after merging latest changes from main.
Still good to go!

PR #110, opened 2024-02-20
Remove JSON Object Length Check

pyth0n1c:
Since a limitation in the API no longer exists,
this length check code is removed.
Testing still generates proper objects with/without enrichment.


ljstella (approved):
LGTM :shipit:

PR #112, opened 2024-02-20
--disable_enrichment to --enable_enrichment

pyth0n1c:
Since disabled enrichment is the more common case, that will now be the default. As such, the command line flag
--disable_enrichment has been changed to --enable_enrichment, with the default setting being written to contentctl.yml being False for all enrichments.


patel-bhavin:
Tested locally and it works!!

patel-bhavin (approved):

PR #115, opened 2024-02-23
Ignore Comments when resolving Macros

pyth0n1c:
A bug was identified by a user where comments in SPL were resolved as macro names.
Validating would fail since we were unable to find these macros.

Now, we use a regex-based solution to remove comment-text in SPL before attempting. 
to resolve all outstanding macros


patel-bhavin:
Tested!

image

current branch fixes this!
image

patel-bhavin (approved):

PR #116, opened 2024-02-24
Enable individual searches by default

pyth0n1c:
Based on a number of community requests, the following
PR allows a Detection to be enabled by default when an app 
is built.
This is accomplished by adding the value
`enabled_by_default: true` 
into a detection.

Please note that ONLY `type: production` searches (not experimental or deprecated)
can be enabled.
Searches that are enabled_by_default must also be `type: TTP`, `type: Anomaly`, or `type: Correlation`. 
They may NOT be `type: Hunting`.
Setting a non-production or Hunting search to enabled_by_default will throw a descriptive error during validation. 

if the key `enabled_by_default` is NOT supplied in the Detection YML, it will default to `false`, maintaining previous behavior.

### DIFF
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -55,6 +55,7 @@ class Detection_Abstract(SecurityContentObject):
     nes_fields: str = None
     providing_technologies: list = None
     runtime: str = None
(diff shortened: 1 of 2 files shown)


patel-bhavin:
Tested after updating the YAML
image
image

patel-bhavin (approved):

PR #117, opened 2024-03-06
Threat object update

patel-bhavin:
If Ip address is an attacker : create a threat object with ip_address
For such cases:
image

New configs :
image


pyth0n1c (approved):
Changes look good!

PR #130, opened 2024-04-05
BA release notes

patel-bhavin:
- Fixes a small bug in story link creation
- Added BA release notes code

image


pyth0n1c (approved):
tested locally on a number of different tags. looks good to me!

PR #132, opened 2024-04-10
Add more fields to savedsearches.conf

pyth0n1c:
Expose more fields in savedsearches.conf


pyth0n1c:
Note that the following appears for a non-deprecated search:
image

And for a deprecated search:
image


This change has also been ported/implemented with `@computed_field` in:
https://github.com/splunk/contentctl/commit/dd4b150491d9c39afd15dd40ed66b96c5bbbcb70

PR #131, opened 2024-04-08
MITRE ATT&CK Enrichment Perf Improvement

ljstella:
This changes what we fetch for ATT&CK to things we actually use. In profiling, this drops the runtime with enrichment enabled significantly.

- `poetry run contentctl -p ../ --enable_enrichment build -t ssa`
  previously: 170 seconds
  after: 40.8 seconds
  get_attack_lookup() before: 165s (97.46% of total runtime)
  get_attack_lookup() after: 36.9s (90.44% of total runtime)

- `poetry run contentctl -p ../ --enable_enrichment build` (ESCU Build)
  full runtime previously: 130 seconds
  full runtime after: 68.8s
  get_attack_lookup() before: 99.6s (76.56% of total runtime)
  get_attack_lookup() after: 36.4s (52.89% of total runtime)

Exact numbers may differ, ymmv, etc. Profiling performed on an i9 Macbook Pro in a relatively warm room.

I think there's still some performance improvements to be found here, but this is a good start.

PR #133, opened 2024-04-11
Porting mitre att&ck perf improvement

ljstella:
See #131 for more details


pyth0n1c (approved):
All changes look good!
Looks like more changes than it actually is as rebase was used to update the source branch.

PR #134, opened 2024-04-12
(Draft) - Generic  Drill down

patel-bhavin:
Based on this feature request - 

https://github.com/splunk/security_content/issues/2385

This PR adds a drill down config to view the specific event for for all TTP detections that have a `risk_object = system` defined in their detection objects.

Example Config:
image

### DIFF
--- a/contentctl/input/detection_builder.py
+++ b/contentctl/input/detection_builder.py
@@ -50,12 +50,32 @@ class DetectionBuilder():
                     self.security_content_obj.deployment = None
                 else:
                     self.security_content_obj.deployment = matched_deployments[-1]
+                    ""
+    def addBasicDrilldown(self) -> None:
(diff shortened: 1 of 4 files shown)


pyth0n1c:
@patel-bhavin I am going to close this out since it was implemented before the contentctl 4.0 updates.  It also introduces a new Pydantic Object called `Drillown`, so that drilldowns can be user-configured instead of just derived from the search.
This superseding PR here with lots of comments from @0xC0FFEEEE which includes many of your ideas in it as well.
https://github.com/splunk/contentctl/pull/256

Let's keep most of our discussion on this new PR. I have linked back here in the comments of that PR so that we can refer to this code/ideas: https://github.com/splunk/contentctl/pull/256#issuecomment-2318842089

PR #135, opened 2024-04-12
Updated GH Actions due to deprecations

ljstella:
Every Actions run is currently surfacing multiple annotations due to deprecated Node actions:

Screenshot 2024-04-12 at 12 39 15 PM

This PR bumps the versions on `.github/workflows/testEndToEnd.yml` and `.github/workflows/pushToPyPiOnRelease.yml` so that these won't be present going forward.


pyth0n1c (approved):
Workflows pass and jobs look good.
Thanks Lou!

Approved.

PR #145, opened 2024-05-14
Add UI dispatch app

pyth0n1c:
~~Add an additional field to savedsearches.conf called~~
~~`request.ui_dispatch_app = SplunkEnterpriseSecuritySuite`~~
~~This ensures proper tracking/accounting for the context under which searches run in-app.~~

The original purpose of this branch has been removed (see `ui_dispatch_app` comments above), but we want to maintain the other infrastructural changes made in this PR.

There are a number of changes to other Jinja2 template files because we now pass the entire app object (from contentctl) into the Jinja2 conf file templates, not just the APP_NAME string.  This gives us more flexibility with what makes it ultimately into the conf file!

We have also cleaned up some imports, typing, and removed some commented-out code blocks.


josehelps:
@pyth0n1c there seems to be quite a few changes outside adding the field on this PR, can you open one without the other commits?

pyth0n1c:
> @pyth0n1c there seems to be quite a few changes outside adding the field on this PR, can you open one without the other commits?

I have removed the original purpose of this PR (see the updated description, we will no longer add `request.ui_dispatch_app` to savedsearches.conf). This PR is now entirely the other structural and hygiene improvements to the codebase.

PR #146, opened 2024-05-15
Enable acs deploy + appinspect warnings

pyth0n1c:
This set of changes adds acs_deploy functionality to contentctl 4.x.
It is mostly a port and cleanup of how it was implemented in contentctl 3.x.

It also addresses some warnings that were present in Cloud Appinspect results.

Note that in order to get these changes in a pre-exisiting contentctl app, you may need to
update files in your app_template folder.
These changes have been made to app.conf, content-version.conf, and default.meta.


contentctl/templates/app_template/metadata/default.meta
pyth0n1c, line 1: Everything below the comment no longer exists as a conf file - should these all be removed from the default.meta file?  I think that is probably the case.

pyth0n1c: We should also evaluate the contents of the distsearch.conf file.

patel-bhavin (approved):

patel-bhavin:
Tested the build locally, works good! This generates all the files for ESCU from contentctl except : `app_template/metadata/default.meta` which we will used from security_content

PR #150, opened 2024-05-20
pin to requests 2.31 to resolve docker library error

pyth0n1c:
This morning, python requests library 2.32 was pushed.
However, this breaks docker-py: https://github.com/docker/docker-py/issues/3256

Since docker-py is a dependency, we must pin to request `~2.31` for now

We also have added setuptools, which is not installed by default in Python 3.12: [Link to Python 3.12 release notes](https://docs.python.org/3/whatsnew/3.12.html#:~:text=not%20pre%2Dinstall-,setuptools,-in%20virtual%20environments)
To pyproject.toml since it is used in pycvesearch: https://github.com/cve-search/PyCVESearch/blob/6a492a72f95ea96f1ef56109d138dfa344d7e9ec/pycvesearch/core.py#L4


patel-bhavin (approved):

ljstella:
requests `2.32.0` and `2.32.1` were both yanked with `2.32.2` being the new release. However, I believe we actually need `docker-py`'s next release which includes a more stable fix. That release appears to be in-progress (https://github.com/docker/docker-py/pull/3262) but is not yet available.

patel-bhavin:
Merging this query and then will create another PR when a more stable fix is out!

Issue #161, opened 2024-06-11
Better error handling when security_content is not present


josehelps:
If user is not running on security_content you get this error, maybe we should generate something more telling that they need a content repo:

```
? enter MITRE ATT&CK Technique IDs related to the detection, comma delimited for multiple T1003.002
? security_domain for detection endpoint
Error during contentctl:
[Errno 2] No such file or directory: '/Users/jhernandez/splunk/contentctl/detections/endpoint'
Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.11/site-packages/contentctl/contentctl.py", line 617, in main
    args.func(args)
  File "/opt/homebrew/lib/python3.11/site-packages/contentctl/contentctl.py", line 359, in new_content
    new_content.execute(new_content_input_dto)
  File "/opt/homebrew/lib/python3.11/site-packages/contentctl/actions/new_content.py", line 23, in execute
    new_content_yml_output.writeObjectNewContent(new_content_generator_output_dto.obj, new_content_generator_output_dto.answers.get("detection_kind",None), input_dto.new_content_generator_input_dto.type)
  File "/opt/homebrew/lib/python3.11/site-packages/contentctl/output/new_content_yml_output.py", line 19, in writeObjectNewContent
    output_folder.mkdir(exist_ok=True)
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pathlib.py", line 1116, in mkdir
    os.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jhernandez/splunk/contentctl/detections/endpoint'
```

ljstella:
Looks like `init` isn't populating the dirs under detections, which the security_domain is mapped to.

PR #162, opened 2024-06-11
fleshing out init dirs

ljstella:
Cleans up #161 

- Adds dirs in `contentctl/templates/detections/` that match the detection kinds we ask about. 
- Set naming to the proper field in the prompts.
- Remove extra fields from detections and stories created with the tool.
- Issue with story usecases
- default testing data link in detection


ljstella:
Sorting outputs is going to be a whole other issue, and can wait until a larger effort is put into `contentctl new` workflow as a whole.

pyth0n1c (approved):
I have tested this locally by creating a new repo with contentctl init and creating a few new detections and stories.
Great work on the PR, thank you!

PR #163, opened 2024-06-13
Removing Bias Language and updating templates

ljstella:
Renamed `[replicationBlacklist]` to `[replicationDenylist]` per documentation.

While working on that, I realized we have hardcoded references to `DA-ESS-ContentUpdate` and `ESCU` in the distsearch.conf template. We should figure a way to remove those as part of this too. Alternatively, we could remove it entirely and let folks decide when to add their own replication denylist by hand, when they need it, covering the files they need.

### DIFF
--- a/contentctl/templates/app_template/default/distsearch.conf
+++ b/contentctl/templates/app_template/default/distsearch.conf
@@ -1,5 +1,0 @@
-[replicationSettings:refineConf]
-replicate.analytic_stories = false
-
-[replicationDenylist]
-excludeESCU = apps[/\\]DA-ESS-ContentUpdate[/\\]lookups[/\\]...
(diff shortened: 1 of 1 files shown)


ljstella:
Confirmed that we don't currently update this value when generating a build from a new repo: 

```
$ cat dist/ContentPack/default/distsearch.conf                                                                                                        
[replicationSettings:refineConf]
replicate.analytic_stories = false

[replicationDenylist]
excludeESCU = apps[/\\]DA-ESS-ContentUpdate[/\\]lookups[/\\]...
```

Removing the file entirely from the app_template so that users: 
1) don't get our config when they don't need it 
2) can ship without a distsearch.conf specific to their content app until they need it 
3) can manage it themselves.

pyth0n1c (approved):
This is a good change. I added a comment with a link for reference to ESCU's distsearch.conf to the PR

PR #164, opened 2024-06-17
Cleaning up output order

ljstella:
1) Removes extra fields that we don't want that were missed before
2) sorts the output! Technically, order doesn't matter in YAML. But given that we're generating a skeleton that we expect people to manually add/edit, we should create a file that's consistent with what we expect.


pyth0n1c (approved):
User interface and output files all look correct.
Of course they still fail validation, but this is **intentional**
as the user needs to update them.

PR #165, opened 2024-06-17
Improve filter macro checking

pyth0n1c:
Improve checking to ensure that a filter macro is included in each detection and that the filter macro has the proper name (which is deterministically calculated from the name of the detection/filename itself).


pyth0n1c:
Tests all look good, including against security_content!

PR #167, opened 2024-06-19
add testing against security_content repo

pyth0n1c:
We want to test against a larger repo of content that is known to be correct.
This can catch more errors with contentctl that we may have missed before.

In this PR, we add a job that checks out the default branch of https://github.com/splunk/security_content
and runs a build, with enrichment, against that content.

Any failures will give us valuable information about potential issues with contentctl itself!


ljstella (approved):
LGTM :shipit:

PR #169, opened 2024-06-20
Add fields as requested

pyth0n1c:
Add new fields to output objects


ljstella (approved):
LGTM

pyth0n1c:
FYI, I am holding this while we do some testing.

PR #170, opened 2024-06-21
Add simple API for a more

pyth0n1c:
Add a simple was to get a contentctl config
and run contentctl operations without going through
the Command Line Interface/main function in contentctl.

Assuming that you are already in a repo that has contentctl.yml, the following should work:
```
from contentctl import api
from contentctl.contentctl import validate_func

#loads by default from ./contentctl.yml, but that can be overridden. 
#config is a dict that overrides default values in that file

cfg = api.configFromFile(config={'enrichments':True})
director_obj = validate_func(cfg)

#director_obj has all your validated, constructed python objects (if you try to interactively print this out, you will get an error. that is expected)
```

### DIFF
--- a/contentctl/api.py
+++ b/contentctl/api.py
@@ -3,7 +3,7 @@ from typing import Any, Union, Type
 #from contentctl import contentctl
 from contentctl.input.yml_reader import YmlReader
 from contentctl.objects.config import test_common, test_servers, test
-
+from contentctl.objects.security_content_object import SecurityContentObject
 
 
 def configFromFile(path:Path=Path("contentctl.yml"), config: dict[str,Any]={}, 
@@ -122,3 +122,18 @@ def updateConfig(config:Union[test,test_servers], **key_value_updates:dict[str,A
     
     return config_copy
     
+
+from contentctl.input.director import DirectorOutputDto
+def contentToDict(director:DirectorOutputDto)->dict[str,list[dict[str,Any]]]:
+    output_dict:dict[str,list[dict[str,Any]]] = {}
+    for contentType in ['detections','stories','baselines','investigations',
+                        'playbooks','macros','lookups','deployments','ssa_detections']:
+        
+        output_dict[contentType] = []
+        t:list[SecurityContentObject] = getattr(director,contentType)
+        
+        for item in t:
+            print(item.model_dump())
+            output_dict[contentType].append(item.model_dump())
+    return output_dict
+            
--- a/contentctl/api.py
+++ b/contentctl/api.py
@@ -1,13 +1,12 @@
 from pathlib import Path
 from typing import Any, Union, Type
-#from contentctl import contentctl
 from contentctl.input.yml_reader import YmlReader
-from contentctl.objects.config import test_common, test_servers, test
+from contentctl.objects.config import test_common, test, test_servers
 from contentctl.objects.security_content_object import SecurityContentObject
+from contentctl.input.director import DirectorOutputDto
 
-
-def configFromFile(path:Path=Path("contentctl.yml"), config: dict[str,Any]={}, 
-                   configType:Type[Union[test,test_servers]]=test)->Union[test,test_servers]:
+def config_from_file(path:Path=Path("contentctl.yml"), config: dict[str,Any]={}, 
+                   configType:Type[Union[test,test_servers]]=test)->test_common:
     
     """
     Fetch a configuration object that can be used for a number of different contentctl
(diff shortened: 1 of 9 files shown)


josehelps:
Nice thank you for the quick turn around. Waiting on CI but LGTM.

josehelps (approved):

contentctl/api.py
```diff
@@ -0,0 +1,124 @@
+from pathlib import Path
+from typing import Any, Union, Type
+#from contentctl import contentctl
```
cmcginley-splunk, line 3: Can this be removed?

contentctl/api.py
```diff
@@ -0,0 +1,124 @@
...
+from contentctl.input.yml_reader import YmlReader
+from contentctl.objects.config import test_common, test_servers, test
```
cmcginley-splunk, line 5: `test_common` is unused

contentctl/api.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+
+def configFromFile(path:Path=Path("contentctl.yml"), config: dict[str,Any]={},
```
cmcginley-splunk, line 9: nit: not something to be addressed in this MR since I know it extends beyond this file, but we should consider going with more traditional PEP8 Python style guidelines and use snake-case for function definitions

contentctl/api.py
cmcginley-splunk: nit: again, not something for this MR, but just calling it out here, but, on the flip-side, class names should not be in snake-case for PEP8

contentctl/api.py
```diff
@@ -0,0 +1,124 @@
...
+    # Force validation of assignment since doing so via arbitrary dict can be error prone
+    # Also, ensure that we do not try to add fields that are not part of the model
+    config_copy.model_config.update({'validate_assignment': True, 'extra': 'forbid'})
+
+    
+    
+    # Collect any errors that may occur
+    errors:list[Exception] = []
+    
+    #We need to do this one by one because the extra:forbid argument does not appear to
```
cmcginley-splunk, line 112: nit: there should be a space after `#`

contentctl/api.py
```diff
@@ -0,0 +1,124 @@
...
+    # The function below will throw its own descriptive exception if it fails
+    configObject = configFromDict(yml_dict, configType=configType)
+
+    return configObject
+
+
+
+
+def configFromDict(config: dict[str,Any]={}, 
+                   configType:Type[Union[test,test_servers]]=test)->Union[test,test_servers]:
```
cmcginley-splunk, line 55: Could the return type not just be `test_common` since `test` and `test_servers` both inherit `test_common`?

cmcginley-splunk:
Could I similarly call `test_common_func(cfg)`?

contentctl/api.py
pyth0n1c: We discussed improving our style guidelines/requirements last week.
During reviews, let's start "requiring" that new functions/classes follow guidelines.
Fixing existing ones to follow those guidelines is a bit more involved, but there's no good reason not to at least start enforcing it for new things :)

cmcginley-splunk: Agreed! Yeah I think if we're onboard to do PEP8 as a style guide, then doing it as we go forward is a good idea :)

contentctl/api.py
pyth0n1c: Yes, this has been removed 👍

contentctl/api.py
pyth0n1c: I removed test_common. You are correct about following Pep-8. I have addressed this for function names in the new file, but have not addressed it for imported classes.
For now, we will get better at new code following proper conventions.  We will eventually have an MR that cleans up existing code as well.

contentctl/api.py
pyth0n1c: Agreed, there were a few other comments lacking this space as well in the file.

contentctl/api.py
pyth0n1c: Yes - both functions that return configs now return a test_common.
We will leave it up to type-checking in the IDE + users to ensure that if they are accessing fields unique to a test or test_servers object that they have requested (and returned) the correct object type.

pyth0n1c:
> Could I similarly call `test_common_func(cfg)`?

Yes, this should work too. I moved the creation of the container objects (part of the `test` object, not the underlying actual containers themselves) from `main` into `test_common_func` to account for this:
https://github.com/splunk/contentctl/blob/bf63cdd76dc8992fc41dfafa20ca8dd8ab3752ad/contentctl/contentctl.py#L105-L108

Doing a bit more logic in the class's pydantic validator functions might be better for this, but we can chat about that.

PR #172, opened 2024-06-25
Release v4.1.0

pyth0n1c:
Collect PRs to merge into main for release 4.1.0
This includes a number of updates: 

- a new workflow is included that tests contentctl by ensuring that the github.com/splunk/security_content repo can be `contentctl build` correctly without errors. Since this content should all be "correct" this is a good test of the contentctl tool
- a simple API is included so that objects can be parsed+loaded and tests can be run without needing to go through the command line interface.  
- API Deploy functionality, which was experimental and allowed individual pieces of content to be deployed to an on-premises instance using the Splunk REST API, has been removed.  This is because it ONLY supported deploying searches and macros. Since we cannot deploy other content, such as lookups or stories, we cannot guarantee that content will work as expected.  This may be recreated with more robust, scalable support at a later date as it is a popular user request.
- Improved `contentctl new` workflow to fix errors and ensure that fields are written in the correct order.
- "CVE Enrichment" now only populates the `url` field of the CVE.  This is used to link directly to the relevant NIST page and supports faster site build time. It also avoids using the CVESearch tool (and the circl.lu API which is frequently down or has extremely slow, multi-minute response times).
- Better filter_macro validation and detection of macros used in searches missing from the macros/ folder


ljstella (approved):
:shipit:

Issue #173, opened 2024-06-25
Add API Deploy Feature


josehelps:
We have a light [implementation](https://github.com/splunk/contentctl/blob/d94aa7ba488a9fa60c1b79faca11514f9392d712/contentctl/actions/api_deploy.py) of how to implement deploy through API but still missing Analytic Stories objects and Lookups through the API.

There was a github PR opened to add auth via token, but it has been closed since we should revisit implementing this feature as a core functionality.

smanaud:
Hello,

We use your nice tool (v3.0.2) and about your future implementation of api_deploy (that it's a very good news), could it be possible to add a secure way to authenticate through splunk ? (instead of username/password/token in clear text in contentctl.yml).
Maybe the possibility to add args via command line to "def api_deploy(args)" like username and password. This will allow us to put github environment variable in our workflow like this :
- name: Deploy the app on Splunk
        env: 
          USER: ${{ secrets.SPLUNK_USERNAME }}
          PWD: ${{ secrets.SPLUNK_PASSWORD }}
        run: contentctl api_deploy -u ${USER} -p ${PWD}

Thank you very much !

PR #175, opened 2024-06-25
Removing rest_deploy, apav_deploy 

josehelps:
They are unused functions in contentctl, I think we should revisit API deployment since it is a requested feature by customers. Here is the tracking github issue: https://github.com/splunk/contentctl/issues/173


pyth0n1c (approved):
I agree with renaming and, for now, removing references to api_deploy.
This is a commonly requested feature, but will take a significant effort to properly build and make robust.  There is an issue tracking this: https://github.com/splunk/contentctl/issues/173
that we will come back to.

PR #180, opened 2024-07-04
Add support for data source objects

P4T12ICK:
Add support for data source objects


contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -369,8 +369,6 @@ def model_post_init(self, ctx:dict[str,Any]):
         # if not isinstance(director,DirectorOutputDto):
         #     raise ValueError("DirectorOutputDto was not passed in context of Detection model_post_init")
         director: Optional[DirectorOutputDto] = ctx.get("output_dto",None)
-        for story in self.tags.analytic_story:
```
pyth0n1c, line 372: Why has this been removed?  This is how we add relevant detections to an analytic story?

pyth0n1c (approved):
PR approved as an initial first past.
We still have some work to do in linking data_source objects directly to the detections that use them (and a few other changes).
Those will come in a future PR.

Great work!

PR #187, opened 2024-07-10
Fix checking during mode:changes

pyth0n1c:
Fix bug where a detection that changed
would not be tested if it contained the
string "dist" in the name.  This was due
to how we checked whether or not
modified content was in the appropriate
content folders like detections, macros,
and lookups or whether it was in a folder
like dist/ or app_template.  The way this
checking is done is now improved so we
examine the paths using is_relative_to
instead of looking for a substring in the
path.


patel-bhavin (approved):

patel-bhavin:
Full path checking, FTW!

PR #188, opened 2024-07-12
Small Model Tweak

ljstella:
Adding the `mitre_attack_id` field back to the detection tags model so that SSE doesn't break- Their move to support the new API did not include support for the `tag.mitre_attack_enrichments[]` object, so sometimes our content just doesn't show up as mapped without this.  

Also tweaked the regex in `mitre_attack_id` because I was getting runtime errors on the use of `\d` to grab digits.

### DIFF
--- a/contentctl/objects/detection_tags.py
+++ b/contentctl/objects/detection_tags.py
@@ -31,7 +31,7 @@ class DetectionTags(BaseModel):
         return round((self.confidence * self.impact)/100)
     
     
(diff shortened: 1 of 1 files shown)


pyth0n1c (approved):
This looks good, thanks Lou!
I have manually run this and verified the updated, correct output in detections.json.
I also confirmed this does not impact the output in the ESCU App itself.

PR #191, opened 2024-07-15
improve output of risk severity field.

pyth0n1c:
it is now calculated using the risk score.

~Note that we may apply this to confidence field as well after discussion~
Given ongoing discussions over the existence of the `confidence` field, we will not make changes to that value in this PR.
Please see this PR for further discussion around the future of `risk stuff` in our YMLs: https://github.com/splunk/contentctl/pull/263


patel-bhavin:
Does this code require the risk_score to be present in the yaml since we are trying to remove risk_score and convert it into a computed field as per this security content PR : https://github.com/splunk/security_content/pull/3062

pyth0n1c:
> Does this code require the risk_score to be present in the yaml since we are trying to remove risk_score and convert it into a computed field as per this security content PR : [splunk/security_content#3062](https://github.com/splunk/security_content/pull/3062)

No, it does not require it to be present.  It USES the risk_score value, but that is a computed_field that is determined at runtime from confidence and impact.

pyth0n1c:
Merging this PR to the release branch.
As that branch has many related changes, they will all be evaluated and merged to main at the same time.

PR #193, opened 2024-07-17
better data source handling

P4T12ICK:
This PR will improve the data source handling. 

Changes:
- data source and event source schema changes
- add data source field to annotations
- create data_source and event_source lookup during build command

Need to be merged before the corresponding security_content PR


contentctl/input/director.py
```diff
@@ -68,17 +67,19 @@ def addContentToDictMappings(self, content: SecurityContentObject):
             # Since SSA detections may have the same name as ESCU detection,
             # for this function we prepend 'SSA ' to the name.
             content_name = f"SSA {content_name}"
-        if content_name in self.name_to_content_map:
+        
+        if content_name in self.name_to_content_map and isinstance(self.name_to_content_map[content_name], type(content)):
```
pyth0n1c, line 71: I am not sure why this change was made.  This check should enforce that no two pieces of content should have the same name (and the check below enforces that no two pieces of content should have the same uuid).  
This modification breaks that because it allows two **different content types** to have the same name, which they should not.

contentctl/input/yml_reader.py
```diff
@@ -40,6 +40,10 @@ def load_file(file_path: pathlib.Path, add_fields=True, STRICT_YML_CHECKING=Fals
         if add_fields == False:
             return yml_obj
         
-        yml_obj['file_path'] = str(file_path)
+        try:
+            yml_obj['file_path'] = str(file_path)
+        except Exception as e:
```
pyth0n1c, line 45: remove this - this is something i added for debugging

contentctl/input/director.py
pyth0n1c: I think I understand now why this change was made - there was a naming confllict in security_content between the osquery MACRO and the osquery DATASOURCE. This has been resolved by renaming the osquery macro to osquery_macro.
Unique names are a key component of contentctl (and how we map from string->content), so this MUST be maintained.

pyth0n1c:
There is a chicken 🐔  and egg 🥚 situation where new contentctl updates are required for security_content to pass and security_content passing requires new contentctl updates. We have performed local testing agains the new security_content updates and they pass - so we accept the CICD failures in contentctl. See the screenshot below for evidence of success:
image

PR #199, opened 2024-07-19
Fix mlmodel changes bug

pyth0n1c:
A modified mlmodel file will cause an exception when mode:changes is run for contentctl test.
This fixes that issue AND provides a more verbose error message when a file besides csv, mlmodel, or yml has been changed in lookups/ directory.  This throws a descriptive exception, only those files should be present in that directory.


patel-bhavin:
Tested with github-inte_schedule branch in gitlab!

patel-bhavin (approved):

PR #200, opened 2024-07-22
Parse and validate CSV files

pyth0n1c:
Ensure that each row has at least
one row, so that fields/headings
can be extracted, and that each
row has the proper number of
fields in it. Report verbose errors
for each row if there is an issue.


pyth0n1c:
Please note this this test is EXPECTED to fail on smoketest_escu since there is currently a badly formatted CSV file in the security_content develop branch.  The smoketest_escu job does fail as expected.

PR #204, opened 2024-07-25
Create threat objects for threat objects

ljstella:
There was a condition in which a field listed as an attacker (which **should** be translated to a Threat Object) was being converted into a Risk Object instead. 

This changes that, so that all observables listed as attackers will be threat objects. 

Because ES creates a risk event for each pair of risk objects and threat objects (1 risk and 3 threats will create 3 risk events, 2 risk and 3 threats will create 6 risk events, etc), this also has the benefit of reducing the total amount of risk events created under our system. 

Relatedly, we no longer need to fail testing if every event in a result set does not have **all** of the fields listed in the observable. The multiplexing, essentially, of raw search results into risk events means that we should only really require each observable to be present _somewhere_ in the dataset. (The change implementing this is not yet in this PR)


ljstella:
Superseded by #234

PR #206, opened 2024-07-26
updating error handling on selected testing

patel-bhavin:
fixing  a bug on selected changes for unit -testing  when selected! 

image


pyth0n1c (dismissed):
This is due to the newly implemented __eq__ method in v4.2.0.  That was added along with __lt__ and __hash__ to make sorting and set operations easier for all SecurityContentObjects.  __eq__ and __lt__ not can ONLY occur between SecurityContentObjects - it should not work (and should raise an exception) for tests between any other objects.
This INCLUDEs None.  In addition since `x is None` is more Pythonic than `x == None`, we probably want to maintain this behavior.

pyth0n1c (approved):
just bumped release version

PR #208, opened 2024-07-27
Forbid extra fields in YMLs

pyth0n1c:
Add an extra, missing field to the lookup.py model called max_matches that was accidentally dropped. set extra='forbid' for SecurityContentObject

This PR is on hold pending significant, structural changes to Detection YMLs with respect to Risk Based Alterting (RBA) content:
https://github.com/splunk/contentctl/pull/263


pyth0n1c:
Note that the `groups` and `context` fields were temporarily added to the detection_tags object and will require some discussion.

PR #210, opened 2024-07-29
SA Admon lookup exclusion

patel-bhavin:
This PR is failing on contentctl build for https://github.com/splunk/security_content/pull/3026 , since there is a detection which this lookup `admon_groups_def` which not present in the ESCU app. 

The lookup is shipped by another app called : `SA-admon`. Hence adding it to this exclusion list!


pyth0n1c (approved):
Looks good!

PR #211, opened 2024-07-30
make labels a bit shorter

pyth0n1c:
We need to prevent the labels for the name of a search from being too long because it can cause issues in the user interface.
As such, we have removed the words `- Experimental -`, `- Deprecated -`, and `- RIR -` from the `action.correlationsearch.label` field in savedsearches.conf.


patel-bhavin:
Tested on Gamorr, the build looks good!  Originally this search errors out when we try to save this. Not any more: 

Slack Thread: https://splunk.slack.com/archives/C06S8S085HP/p1722267799266179

Testing:

image

image

patel-bhavin (approved):

PR #212, opened 2024-07-30
Add a launcher to contentctl.py to allow easier debugging and launchi…

Res260:
Add a launcher to `contentctl.py` to allow easier debugging and launching with `python -m contentctl.contentctl`

### DIFF
--- a/contentctl/contentctl.py
+++ b/contentctl/contentctl.py
@@ -229,4 +229,7 @@ def main():
             print(e)
             
         sys.exit(1)
-    
+
+
+if __name__ == "__main__":
+    main()


pyth0n1c (dismissed):
Yes, this looks like an improvement for some types of testing/debugging without doing a poetry install or pip install :)

pyth0n1c:
For awareness of anything else looking at the PR, this addition means that you can do things like
```
python -m contentctl.contentctl validate
python -m contentctl.contentctl test --post-test-behavior always_pause mode:all
```
which is convenient, because you do not need to `poetry install / poetry shell` or `pip install contentctl` to run the tool as long as you have the requirements met.

Res260:
Yes and it also allows to put breakpoints in IDEs like pycharm and VS code (idk if there is another way but its usually how I do it haha)

PR #213, opened 2024-07-31
Fix name length

pyth0n1c:


pyth0n1c:
Requires an update to where we use `savedseaches://` construction to reference a savedsearch, such as in savedsearches_analyticstories.j2

pyth0n1c:
Merging this into the release branch.
That whole branch will be reviewed as one large PR as it has significant changes around some of the same parts/fields.

PR #216, opened 2024-08-05
Add latest TA version validation

P4T12ICK:
Add a new parameter to validate data_source_TA_validation which allows the validation of latest TA version in data source.


pyth0n1c (approved):
I have run this locally and, after the latest TA versions were updated in the data_source objects in security content, this is passing CI/CD!
Approved.

PR #217, opened 2024-08-05
Allow `contentctl test` to work on Windows by fixing a path problem.

Res260:
When using `contentctl test` on Windows, a wrong path is generated. We need to force this path to be a posix path else we have an error that looks like this:

```
400 Client Error for http+docker://localnpipe/v1.46/containers/create?name=contentctl_0&platform=linux%2Famd64: Bad Request ("invalid mount config for type "bind": invalid mount path: 'C:/tmp/apps' mount path must be absolute"
```

This should not have an impact on Linux systems, but I have not tested myself.

### DIFF
--- a/.github/workflows/testEndToEnd.yml
+++ b/.github/workflows/testEndToEnd.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.11", "3.12"]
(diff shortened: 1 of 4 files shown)


pyth0n1c (approved):
This looks good and still works in CICD under linux and on my macos machine.
I have added a windows test runner to the operating_system matrix for ongoing testing.

pyth0n1c:
Thank you for the PR @Res260 ! We do not have windows machines, so were not aware this was an issue!

PR #219, opened 2024-08-08
Remove SSA specific code

P4T12ICK:
Remove code related to ssa detections


pyth0n1c (approved):
During my review, I encountered some issues where build_ssa and referneces to it were not removed. Those have been fixed.
Similarly, detection.search was still typed as a dict|str (to support sigma content) and all references to that have been converted to JUST a str.
Looks good and I approve!

PR #235, opened 2024-08-13
Update on naming for the repo readme vs app readme

pyth0n1c:
Repo and app readme are are actually
different. Also make any usage of the slim
packaging feature generate an exception.


pyth0n1c (commented):
These changes work in CICD as well as locally. Merging now!

PR #239, opened 2024-08-15
add publish_date field

pyth0n1c:
add publish_date field to action.correlationsearch.metadata

Date should be the epoch time of Detection.date @ midnight, UTC. It should be a whole number and expressed as a float.

image


pyth0n1c:
I have confirmed with another Splunker that these changes are correct and will merge them.

PR #240, opened 2024-08-15
Hotfix: Bumping integration testing timeout to compensate for recent bugfix

cmcginley-splunk:
# Context
- in #92 we introduced a bugfix which shaved an unnecessary extra 60s wait before starting the backoff timer
- this resulted in less overall time for those detections which are sluggish to generate risk events, causing some intermittent failures

# Code changes
- we bumped the total backoff timer by 60s (from 210s to 270s)


pyth0n1c (approved):
Agreed with this change. This amount gives us a bit larger margin and should cut our failures to 0. If we continue to see timing issues, we can bump this even higher.

PR #241, opened 2024-08-15
Enabling risk/observable matching

cmcginley-splunk:
# Context
* When risk message validation was added, an early version of risk/observable matching was added as well, but kept disabled

# Code changes
* Risk/observable matching enabled
* Risks are matched against observables using the field extracted from `contributing_events_search`
* The following validations are performed
  * All risk events can be matched to an observable
  * All observables are matched to at least one risk event
  * All risk events have the expected typing relative to matched observable
  * All risk events can be matched to only one observable
  * All risk events match observables with valid roles only (e.g. no risk events match `Attacker` role observables) 
    * There is one caveat w/ this validation
    * It's possible, based on timing, that a detection may pass this validation falsely
    * If at the time of checking, the offending risk event matching the Attacker observable has not yet been generated and all other risk events pass validation, then this condition will seemingly be met
    * See #248 for a possible future mitigation for this to ensure this validation is applied more reliably

# TODO
- [ ] Remove comment (or annotate w/ GitHub issue) relating to having multiple conflicting roles once I get feedback
- [ ] Deploy to security_content and security-content-automation after merge
- [x] Disable verbose logging

# Testing
* Preliminary testing has shown the following true failures:
  * ~50 detections have Attacker role observables that can be matched risk events incorrectly
    * **NOTE**: Lou's PR #234 has been shown to resolve these issues
  * 7 detections show risk events with improper types
    * 6 of these are due to an issue in `contentctl build` which causes `User Name` observables to create `other` type risk events instead of `user` type (see #246)
    * The other is also due to issues in `contentctl build`, which causes any observable with a role of `Other` (regardless of its type) to create `other` type risk events (see #247)
  * 2 detections have an observable repeated by mistake
* Actionable fixes needed in ESCU are being tracked via internal tickets

# Future Work
* ESCU content fixers (tracked internally)
  * repeated observable blocks
  * risk events with type other instead of user
  * observables creating risk events instead of threat objects
* #246 
* #247 
* #248 
* #249 
* #250 
* #251 
* #252


contentctl/objects/risk_event.py
```diff
@@ -1,17 +1,55 @@
...
+# TODO (cmcginley): validation of threat objects
+# TODO (cmcginley): 'Email' -> other is where it goes in practice, but I feel like it SHOULD go to
+#   user?
+# TODO (cmcginley): Should we map some of these others more specifically? E.g. 'Process' to
+#   'host_artifacts' and 'File Hash' to 'hash_values'
+# TODO (cmcginley): centralize this mapping w/ usage of SES_OBSERVABLE_TYPE_MAPPING (see
+#   observable.py) and the ad hoc mapping made in detection_abstract.py (see the risk property func)
+
 # TODO (PEX-433): use SES_OBSERVABLE_TYPE_MAPPING
 TYPE_MAP: dict[str, list[str]] = {
```
cmcginley-splunk, line 20: Should we move some of the observable types into more appropriate mappings? e.g. `File Hash` to `hash_values`

contentctl/objects/risk_event.py
cmcginley-splunk, line 1: Should all roles that aren't `Attacker` be generating risks?

E.g. `Victim`, `Child Process`, `Other`, `Target`, `Unknown`, etc.

cmcginley-splunk, line 1: Observable roles are a list not a single value; do we have validation that would prevent conflicting combinations? For example, Attacker paired with any other role

pyth0n1c: Per the following PR, there should ONLY be Attacker and Victim roles, nothing else.  This will be enforced programmatically: https://github.com/splunk/contentctl/pull/243/files

pyth0n1c: These observables should be single-valued and not lists (even though they are typed as lists today). The length of this list is now forced to be one per the following PR: https://github.com/splunk/contentctl/pull/243/files#diff-aeb40da5f7c944d37ef6585996299734dfd6c65b5829e636c715738e5a05a720R29

This decision was made in the interest of minimal changes to the code.  Observables will soon be replaced by (probably) separate Risk Object and Threat Object Objects (rather than a generic observable).

contentctl/objects/risk_event.py
pyth0n1c: `File Hash` under `hash_values` makes sense to me.  There are a few that I am a little less certain about. 
is `MAC Address` best as a system (IP Address is a system artifact, but MAC address is even more closely tied to a system than that), network_artifact, host_artifact, or other?
Lots of other things here, like why shouldn't registry stuff be a host_artifact? I will reach out to Lou and get his thoughts and link to this comment :)

contentctl/objects/risk_event.py
pyth0n1c: As an additional note (this isn't what you're getting at with your comment, but it reminded me), we have MANY fields which are typed as `lists` which should really be typed as `sets`.

contentctl/objects/risk_event.py
```diff
@@ -59,6 +78,10 @@ class RiskEvent(BaseModel):
         default=[]
     )
 
+    # Contributing events search query (we use this to derive the corresponding field from the
+    # observables)
+    contributing_events_search: str
+
     # Private attribute caching the observable this RiskEvent is mapped to
     _matched_observable: Optional[Observable] = PrivateAttr(default=None)
```
pyth0n1c, line 86: Nit: update typing to `Observable | None` in line with new style guidelines

pyth0n1c (commented):
Initial static review with responses to RFC and followup comments. I will merge Lou's PR, update the branch, then do some runtime testing of the PR.
As such, this is not approved YET.

contentctl/objects/risk_event.py
cmcginley-splunk: Cool! I'm gonna make this a follow up issue though for now, so it doesn't block this PR

cmcginley-splunk: #259

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -322,12 +322,13 @@ def nes_fields(self) -> Optional[str]:
...
         return ProvidingTechnology.getProvidingTechFromSearch(self.search)
-        
-    
+
+    # TODO (#247): Refactor the risk property of detection_abstract
     @computed_field
     @property
     def risk(self) -> list[dict[str, Any]]:
         risk_objects: list[dict[str, str | int]] = []
+        # TODO (#246): "User Name" type should map to a "user" risk object and not "other"
```
pyth0n1c, line 331: Why not make this change now instead of tracking it as a TODO?


Second, (I realize you didn't touch these values, but still think this is a valuable comment for the whole team):
We have lots of "magic enums/lists/values" that have come from... somewhere. We should link to documentation when possible so we have a record of why they were chosen (and can update them, if required).  These are all over the place in contentctl - if there is public documentation on where this list of types came from, we should:
1. include it as a comment
2. Break it out into an enum or class

contentctl/objects/detection_tags.py
```diff
@@ -52,6 +52,8 @@ def risk_score(self) -> int:
 
     mitre_attack_id: List[MITRE_ATTACK_ID_TYPE] = []
     nist: list[NistCategory] = []
+
+    # TODO (#249): Add pydantic validator to ensure observables are unique within a detection
```
pyth0n1c, line 56: Nearly all of our `list[ALMOST_ANY_TYPE]` typing should exhibit this uniqueness constraint. We have a few options here:
1. Convert as many things as possible to sets (`SecurityContentObject_Abstract` has the appropriate methods for object comparisons, but we have list[] of things that are not `SecurityContentObject_Abstract` as well). For example, a detection should not be able to have the same AnalyticStory twice.  However, the issue here is we would want duplicates to throw an Exception, not just appear once in the results. Additionally, the orders of some of our serialization are important (so we would need to make sure when they are written to output files, the values are sorted).
2. Define a custom type, like `listWithNoDuplicates[ANY_TYPE]` which ensures no duplicates exist during validation and throws a ValueError on the field if they do.
3. Something else I haven't thought of.

This is out of scope for this PR, but a good point to discuss.

contentctl/objects/risk_event.py
```diff
@@ -1,32 +1,51 @@
...
+        "File Hash",
+        "Resource UID",
+        "Uniform Resource Locator",
+        "File",
+        "Geo Location",
+        "Container",
+        "Registry Key",
+        "Registry Value",
+        "Other"
+    ]
```
pyth0n1c, line 44: Is there a reason that we have not unified this object with the following, in one single place?
https://github.com/splunk/contentctl/pull/241/files#diff-9e1935fe7233e929a9fe50af50183ca0a44b882b79e198b8fbe09ee75c00b55dR332
(other than the fact that what is linked above is more legacy code)?

If these SHOULD be the same, then both parts of the code should refer to one, consistent object.

contentctl/objects/abstract_security_content_objects/detection_abstract.py
cmcginley-splunk: Will make that change if you agree then!

Also +1 on the magic enums. We also don't use the enums consistently. For example, we use the observable type enums in validation, but we have a separate static list that we use in risk generation. Unifying those is a good idea

contentctl/objects/detection_tags.py
cmcginley-splunk: Agreed w/ converting as many things to sets as possible

Re: ensuring duplicates are not added to the set, I would just define a validator as part of our base model and apply it to each field we want it on directly

contentctl/objects/risk_event.py
cmcginley-splunk: We certainly can and should unify these and I have captured the intent to do so in an issue already: https://github.com/splunk/contentctl/issues/247

Doing so in this PR felt out of scope (trying to avoid scope creep)

contentctl/objects/abstract_security_content_objects/detection_abstract.py
cmcginley-splunk: Actually, addressing this here feels a bit like scope creep as it is not part of the feature I am delivering. I also believe @ljstella may already be tracking this change?

pyth0n1c: Agreed as out of scope. Let's lean on the @ljstella updates for this.

contentctl/objects/risk_event.py
pyth0n1c: agreed that this is out of scope. resolved

pyth0n1c (approved):
approved after discussions with dev through comments and slack

PR #243, opened 2024-08-19
New observable role enum

ljstella:
This replaces uses of `SES_OBSERVABLE_ROLE_MAPPING` with a new `RBA_OBSERVABLE_ROLE_MAPPING` for splunkd content (left SSA alone for now)- this new enum is a subset of the old one, only allowing for two possible values. I've modified the existing validator to use this, as well as an additional validation that limits tags.observable[].role to a length of 1. So now, every observable will be *either* an Attacker (threat object), or a Victim (risk object), and we can begin to transition to a new `rba` object.

Content changes because of this change, including updates to deprecated detections, are in this PR: https://github.com/splunk/security_content/pull/3061

I've also updated the templated detection to reflect this new behavior. Smoketest jobs will fail until the linked security_content PR is merged.


pyth0n1c (approved):
These code changes look good. I am approving this, but we should get a full integraiton test of our content under our belt before actually merging the changes and releasing a new version of contentctl.

Note that failing ESCU Smoketest Testing Jobs is allowed because the pending PR with security_content changes has not been merged yet.

PR #255, opened 2024-08-22
Removal of more bits of SSA

ljstella:
Some more SSA stuff to remove.


ljstella:
Only sections left _appear_ to be in release_notes.py - not 100% that those have a replacement just yet.

pyth0n1c (approved):
I independently looked for more SSA bits in the code as well and, other than what @ljstella pointed out in his comment did not find any.
Looks good to me, thank you for the cleanup!

pyth0n1c:
> Only sections left _appear_ to be in release_notes.py - not 100% that those have a replacement just yet.

I agree. I saw these in there and we chatted offline about the one additional reference and it was removed 👍

PR #271, opened 2024-08-28
Abstract Commonly Used Annotated Type Definitions

pyth0n1c:
Add annotated type definitions for commonly used Annotated Strings. This avoid redefining them in multiple places.
It also ensures that if we need to make an update, it only happens in one place.

The MITRE CVE Type annotations were wrong in a number of places - because they were not defined correctly, (notable, they were missing the `Field()` around the pattern) they were not working. For example:
`id: Annotated[str, r"^CVE-[1|2]\d{3}-\d+$"]` is WRONG and should be 
`id: Annotated[str, Field(pattern=r"^CVE-[1|2]\d{3}-\d+$")]` if used directly


ljstella:
@pyth0n1c none of the CI tests run with `--enrichments` - have you tested running this with that?

ljstella (approved):
LGTM

PR #274, opened 2024-08-29
Improve lookup regex - Step 1 - ESCU 5.0

pyth0n1c:
Significantly improve the lookup regex. Allow inclusion of a number of different arguments to lookup and introduce validations of the inputlookup and outputlookup commands as well.


ljstella (approved):
LGTM- good to merge and clean up later

PR #276, opened 2024-09-05
Initial implementation of updating offline caching and loading

pyth0n1c:
Big overhaul here - but need to update the serialization of these objects as the json api objects are different and far too large.
We need to decide if we want this level of complexity for these objects, which is especially nice if we will ever use other fields for these objects.

If this complexity is to much, we can also use `extras="ignore"` for the mode and only create the fields that we actually care about.

This is a draft as output api objects are still incorrect (they are far too large and include all the extra fields, so the serialization logic must be updated).


patel-bhavin (dismissed):

pyth0n1c:
Closing this PR in favor of a simpler one with less changes here: https://github.com/splunk/contentctl/pull/291

PR #278, opened 2024-09-06
Fix unintended whitespace

pyth0n1c:
Trailing newline(s) in string fields may cause issues with some parsers when written to conf file. In particular, it causes issues with this open source splunk conf file parser: https://github.com/splunk/addonfactory-splunk-conf-parser-lib
Note that this issue does NOT occur in the Splunk product itself.

This PR removes leading AND trailing whitespace when long string fields are serialized to conf file.


contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -74,6 +74,8 @@ class Detection_Abstract(SecurityContentObject):
 
     data_source_objects: list[DataSource] = []
 
+
```
cmcginley-splunk, line 77: nit: functions w/in a class should only have 1 empty line between them

cmcginley-splunk (approved):
Confirmed that this fix does indeed resolve any in product parsing errors

contentctl/objects/abstract_security_content_objects/detection_abstract.py
pyth0n1c: Agreed, fixed and pushed.
Resolving

PR #254, opened 2024-08-22
First crack at default config for `ruff` - Step 3.5 - ESCU 5.0

ljstella:
Creating first rough draft PR after conversation w/ Eric about getting this started in a new branch.

Rough draft of the default settings and whatnot for implementing `ruff` as a linter & formatter. We'll need to take some time to go through the formatting, and the lints, and see what we do and don't want to turn on.

TODO:

- [ ] Confirm current settings, making changes if required
- [x] Add github ci workflow to format & lint (without applying fixes, and failing on jobs that don't pass?)
- [x] Add precommit hook

PR #289, opened 2024-09-14
initial pass at resolving ruff issues

pyth0n1c:
Tons of changes, many of them small (like resolving unused imports or variables) but great code cleanup.
This effort is motivated by the following PR to add `ruff` support in CI/CD: https://github.com/splunk/contentctl/pull/254

However, because there are so many changes, maybe they should be made after the PR backlog has been cleared out?

This does not begin to address formatting issues, just things like bad imports, unused variables, etc.


pyth0n1c:
> I would agree- clearing the PR backlog first will make this easier so we won't have to spend even more time potentially clearing conflicts in every open PR.
> 
> I am open to doing this in two stages where we enforce the linter first and then fast-follow w/ the formatter if that means getting the first one of these changes (this one) in faster.

I agree.
The takeaway is that this PR only took maybe 30 minutes to throw together, so the initial pass of changes will be pretty easier!
Easier than keeping this PR up to date as we merge changes. I am closing this out - we will reopen a PR later with all the changes rather than continuously update this one.

PR #233, opened 2024-08-08
Add optional explanation field to detections

ryanplasma:
The current behavior of contentctl copies the detection description to both the description and explanation fields in the analytic story page.

This PR adds the option to have an `explanation` field in the detection yml file if one wants to have different content in that area, otherwise it will defer to the content in the `description` field.

This could be expanded on and included in the required fields in the future, but this felt like a good first step.

Image below illustrates the different sections I'm talking about.

image

PR #296, opened 2024-09-26
Ryanplasma add explanation

pyth0n1c:
This PR has been copied from the following external pr contributed by @ryanplasma:
https://github.com/splunk/contentctl/pull/233

This change includes some more documentation on the Explanation field.
Additionally, it fixes a bug where the explanation field is defined, newlines in it are 
not properly escaped.

Since I am unable to edit the original PR, I am closing that out and linking to this one
with changes.


pyth0n1c (commented):
I have tested this and checked that the output is correct

PR #297, opened 2024-09-26
Fix datasource in contentctl new

pyth0n1c:
Remove erroneous spaces from datasources used by `contentctl new --type detection`. Also remove the list of datasource from new_content_questions and instead use the datasource defined in enums and convert them to a list


pyth0n1c (commented):
these changes are straightforward, so i will approve

PR #299, opened 2024-09-30
Option to not fail on missing detection during metadata check

pyth0n1c:
add new option to contentctl inspect that makes missing detections NTT an error - instead just print a warning message.

### DIFF
--- a/contentctl/actions/inspect.py
+++ b/contentctl/actions/inspect.py
@@ -297,10 +297,10 @@ class Inspect:
             validation_errors[rule_name] = []
             # No detections should be removed from build to build
             if rule_name not in current_build_conf.detection_stanzas:
-                if config.exception_on_removed_detections:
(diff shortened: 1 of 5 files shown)


patel-bhavin:
Tested this : suppression worked correctly when added to the CLI however, when I do not add that flag, i should be getting errors and I did not.
Here's the command that I used:

```
contentctl inspect --splunk-api-username "$APPINSPECTUSERNAME" --splunk-api-password "$APPINSPECTPASSWORD" --stack_type victoria --enrichments --enable-metadata-validation --supress_missing_content_exceptions
```

and lets use hypens in the flag name to keep it consistent : update to  use `--suppress-missing-content-exceptions`

minor typo in supress

pyth0n1c:
closing this PR out as is was superceded by https://github.com/splunk/contentctl/pull/305

PR #300, opened 2024-10-04
Exception on malformatted unit tests in YMLs

pyth0n1c:
This PR includes two additional fixes:
1. improve regex and logic used to search for macros in SPL.  Previously, a comment with a newline in it (or other edge cases) could be parsed as a macro when it should have been parsed as a comment.  This resulted in some comments appearing to be "macros not found in contentctl macros/ folder"
2. When testing under --mode changes, an individual piece of content could be erroneously.tested multiple times if it met several conditions. Logic has been updated (by tracking changes in a set rather than a list) so that content cannot be added more than once to the group of content that must be tested in a single test run.
- It was new content not in the target branch
- It had been changed locally, but the changes had not yet been committed.

Raise exception on parse of unittest from yml. 
Do this rather than trying to convert it into an integrationtest or manualtest.
For example, the following YML section is NOT if the file test_data.log does not exist:

```
tests:
- name: True Positive Test Which is not valid but fails silently
  attack_data:
  - data: data/data_file_that_does_not_exist.log
    sourcetype: anySourcetype
    source: anySource.log
```

Because the typing of detection.tests is as follows (since additional tests will be created at runtime):
https://github.com/splunk/contentctl/blob/5488ca68ccfd61b01a58dcd9ef82a38a7eeb61a8/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L88

if it fails to parse from YML as UnitTest, it will then try to parse as the other types (which may pass - but we do not want this behavior).  
https://github.com/splunk/contentctl/blob/main/contentctl/objects/unit_test.py

For example, this will construct a VALID Integration Test (which is not desired behavior):
https://github.com/splunk/contentctl/blob/main/contentctl/objects/integration_test.py

This PR cause a ValueError to be generated if the data read from the YML cannot be used to create a valid UnitTest.


pyth0n1c (commented):
These are fairly minor changes - I have reviewed them and they look good!

PR #301, opened 2024-10-07
Testing on Datasource changes

ljstella:
When we calculate changes between two branches in order to run tests, we currently do not catch data source object changes as content dependencies. Following the same path in which we calculate macros and lookups as being content dependencies, this PR will attempt to change the diff calculation to include data source object changes. 

Notably, this will then trigger the testing for each of these detections when we have an "Auto Update TA" PR like so: https://github.com/splunk/security_content/pull/3154

This testing will provide the confidence and verification that such changes can be merged without issues.

### DIFF
--- a/contentctl/actions/detection_testing/GitService.py
+++ b/contentctl/actions/detection_testing/GitService.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
 from contentctl.objects.macro import Macro
 from contentctl.objects.lookup import Lookup
 from contentctl.objects.detection import Detection
+from contentctl.objects.data_source import DataSource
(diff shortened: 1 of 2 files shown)


ljstella:
Update: https://github.com/splunk/security_content/pull/3197 should handle updates to contentctl which gives us a path forward here.

pyth0n1c (approved):
Tested these changes locally with a number of changes to detections and data_sources.  It looks like it picks up those changes correctly, this looks good to me!

pyth0n1c (approved):
Approved again. I asked the owner to update the version in prep for a contentctl release and it has been done.

PR #303, opened 2024-10-08
Feature: validation of detections against cms_main

cmcginley-splunk:
# Context
* We need to validate that our detections not only work in ES, but are also being parsed correctly by the content versioning service

# Code change
* Added `ContentVersioningService`
  * Enables content versioning
  * Activates CMS parser
  * Validates detections against CMS events, reports any missing detections or other validation issues. Validations include:
    * Counts between `cms_main` and detections based on YAMLs match
    * Each detection can be matched to a `cms_event`
    * Each `cms_event` can be matched to a detection
    * No `cms_event` is repeated
    * UUIDs match between cms/detection pairs
    * Versions match between cms/detection pairs
    * Full search name matches between cms/detection pairs
    * Correlationsearch label matches between cms/detection pairs
  * Returns a list of setup functions which are optionally inserted into the main setup of `DetectionTestingInfrastructure`
* Aggregating errors across all threads (setup, testing, view, view shutdown) and reporting en masse
* Broke up setup functions list to allow for optional inclusion of the content versioning setup functions
* **NOTE**: Now we `raise` any exceptions that crop up during setup, instead of just logging them
* Added logic to `DetectionTestingInfrastructure` to query installed ES version, so we can determine when it is appropriate to run content versioning validation
  * Content versioning validation is only run when 1) integration testing is enabled, 2) ES 8.0.0+ is installed, and 3) when the testing mode is `All`
* Moved logging utility in `CorrelationSearch` (useful for debugging) to `helper.utils.Utils` so it could be re-used in `ContentVersioningService`

# Testing
* I've tested locally w/ some quick and dirty edits to the `savedsearches.conf` file and all seems to work well
* Ran against `escu-production` and all results passed (see attached truncated log)

Screenshot 2024-10-08 at 12 38 55 PM

[content_versioning_service.log](https://github.com/user-attachments/files/18067851/content_versioning_service.log)

# Caveats
* Note that we now raise exceptions from setup, instead of logging and dropping them; I believe this is desirable
* I went back and forth as to whether this validation should happen as part of setup, OR if I should create some new type of test case. I opted for the former as this type of testing doesn't fit our current testing paradigm; validation against the `cms_main` index should happen in aggregate, a test of the whole package, not individual detections. We do not have an easy way to insert testing at that level presently.
* Additionally, this validation *probably* only needs to be run against a single instance and not necessarily all instances in a single run, but I don't see an easy or clean way to do that as well, and there's no harm or impact on execution time as far as I can tell

# TODOs
- [x] disable logging
- [x] remove TODOs or create follow up issues
- [x] Test in CI when able


contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,548 @@
...
+#   - [ ] validate multi-line fields -> search, description, action.notable.param.rule_description,
+#         action.notable.param.drilldown_searches
+
+# TODO (cmcginley): suppress logging
+# Suppress logging by default; enable for local testing
+ENABLE_LOGGING = True
+LOG_LEVEL = logging.DEBUG
+LOG_PATH = "content_versioning_service.log"
+
+# TODO (cmcginley): would it be better for this to only run on one instance? Or to consolidate
```
cmcginley-splunk, line 26: Currently, validation of `cms_main` occurs as part of the instance setup for every instance during integration testing. This is not strictly necessary, as, if versioning is enabled properly on one instances, it presumably is setup appropriately on all other instances. We don't lose anything by running this validation in parallel, as we want to wait to begin testing until this step is complete whether it's being run on one instance or 40.

Any thoughts @pyth0n1c ?

contentctl/actions/detection_testing/DetectionTestingManager.py
```diff
@@ -65,19 +65,27 @@ def sigint_handler(signum, frame):
...
-        
+
         with concurrent.futures.ThreadPoolExecutor(
             max_workers=len(self.input_dto.config.test_instances),
         ) as instance_pool, concurrent.futures.ThreadPoolExecutor(
             max_workers=len(self.input_dto.views)
         ) as view_runner, concurrent.futures.ThreadPoolExecutor(
             max_workers=len(self.input_dto.config.test_instances),
         ) as view_shutdowner:
+            # Capture any errors for reporting at the end after all threads have been gathered
```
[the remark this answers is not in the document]
pyth0n1c: Looks good, this is much better than we had before.
On a slightly related notes, I don't really like using futures as I find them hard to maintain and debug.
These are an improvement over what we had before, which was just a bunch of threads we managed manually.
If we ever have the time and motivation, I'd love to go over how to update this workflow entirely.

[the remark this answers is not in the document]
cmcginley-splunk: Yeah I'd love to discuss that as well!

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -147,15 +164,86 @@ def setup(self):
                 self.check_for_teardown()
 
         except Exception as e:
-            self.pbar.write(str(e))
+            msg = f"[{self.get_name()}]: {str(e)}"
             self.finish()
-            return
+            if isinstance(e, ExceptionGroup):
+                raise ExceptionGroup(msg, e.exceptions) from e                                      # type: ignore
+            raise Exception(msg) from e
```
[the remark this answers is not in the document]
pyth0n1c: Much better

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -269,15 +357,15 @@ def configure_imported_roles(
...
-            # Set which roles should be configured. For Enterprise Security/Integration Testing,
-            # we must add some extra foles.
-            if self.global_config.enable_integration_testing:
-                roles = imported_roles + enterprise_security_roles
-            else:
-                roles = imported_roles
 
+        # Set which roles should be configured. For Enterprise Security/Integration Testing,
+        # we must add some extra foles.
+        if self.global_config.enable_integration_testing:
```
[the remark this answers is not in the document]
pyth0n1c: I agree. I bet this second POST is the result of a copy/paste error (and there by mistkae)

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -1401,6 +1482,7 @@ def hec_raw_replay(
     def status(self):
         pass
 
+    # TODO (cmcginley): the finish function doesn't actually stop execution
     def finish(self):
```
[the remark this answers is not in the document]
pyth0n1c: It does do some extra cleanup, depending on whether it is running using containers or servers:

Removes the container when running using `containers`
https://github.com/splunk/contentctl/blob/98573b0fcc6040bc10046bf444c29d2685545ba6/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructureContainer.py#L29-L36


Nothing extra when running using `servers`: https://github.com/splunk/contentctl/blob/98573b0fcc6040bc10046bf444c29d2685545ba6/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructureServer.py#L10-L11

contentctl/actions/detection_testing/views/DetectionTestingViewCLI.py
```diff
@@ -51,6 +51,8 @@ def showStatus(self, interval: int = 1):
         while True:
             summary = self.getSummaryObject()
 
+            # TODO (cmcginley): there's a 1-off error here I think (we show one more than we
+            #   actually have during testing)
             total = len(
```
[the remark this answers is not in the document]
pyth0n1c: This has been a longstanding issue that. I think, was initially added to avoid implementing lots of logic to avoid divide-by-zero errors.
Let's fix this once and for all as it's been an annoyance for a long time.

contentctl/output/conf_output.py
```diff
@@ -72,7 +72,9 @@ def writeAppConf(self)->set[pathlib.Path]:
                                               [self.config.app]))
         return written_files
 
-        
+    # TODO (cmcginley): we could have a discrepancy between detections tested and those delivered
+    #   based on the jinja2 template
+    #   {% if (detection.type == 'TTP' or detection.type == 'Anomaly' or detection.type == 'Hunting' or detection.type == 'Correlation') %}
     def writeObjects(self, objects: list, type: SecurityContentType = None) -> set[pathlib.Path]:
```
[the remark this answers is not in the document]
pyth0n1c: We do this type of thing A LOT, including in all the Pydantic Validations.
The initial idea here is to treat Detections differently than Baselines, as you can see in the Jinja2 templates:
https://github.com/splunk/contentctl/blob/390c3727bf83b5af3e50e4ed4434b542a7d8629f/contentctl/output/templates/savedsearches_detections.j2
https://github.com/splunk/contentctl/blob/390c3727bf83b5af3e50e4ed4434b542a7d8629f/contentctl/output/templates/savedsearches_baselines.j2#L6

This comes from a time when a Baseline and a Detection were defined as the same object, I believe.

Let's talk more about how to actually fix this at scale.  I also don't like how Baselines and Detections have SO MANY fields in common, but they are totally different objects (that only inherity from SecurityContentObject).

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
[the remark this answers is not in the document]
cmcginley-splunk: Should this update the sync_obj.terminate field though?

contentctl/actions/detection_testing/views/DetectionTestingViewCLI.py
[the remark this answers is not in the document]
cmcginley-splunk: https://github.com/splunk/contentctl/issues/338

contentctl/output/conf_output.py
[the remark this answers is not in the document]
cmcginley-splunk: https://github.com/splunk/contentctl/issues/339

contentctl/objects/content_versioning_service.py
pyth0n1c: I am fine with doing this on every instance.  Right now we do make the very intentional assumption that all instances have the same versions of everything (apps. splunk core, etc). However, I could see in the future setting up multiple different versions configurations with test_servers and wanting to run validation on all of them.

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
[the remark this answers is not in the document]
pyth0n1c: The _INTENT_ of the `.terminate` field is to indicate to all  running futures (whether they are doing container setup, execution, or shutdown) that they should stop early.   The `futures.as_completed` calls are SUPPOSED to catch any exceptions returned from the individual runners, for example:
https://github.com/splunk/contentctl/blob/8362af754acf327ec00daeb4bd5e189ae5722a84/contentctl/actions/detection_testing/DetectionTestingManager.py#L102
And set `.terminate = True` where appropriate.  So we shouldn't have to do this in the finish() function, or elsewhere, if an exception is raised.

But at the same time, I'm not extremely confident that we haven't missed some edge case where we shut down early, but DON'T raise an exception which is caught by the `future.result()` calls:
https://github.com/splunk/contentctl/blob/8362af754acf327ec00daeb4bd5e189ae5722a84/contentctl/actions/detection_testing/DetectionTestingManager.py#L100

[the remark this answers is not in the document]
cmcginley-splunk: Gotcha. Yeah so that was exactly the issue that this PR fixes. Previously, errors were suppressed in setup, so even if setup failed, we would proceed to testing

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,530 @@
...
+            raise ValueError(f"Unable to parse response body as JSON: {e}") from e
+
+        # Find the versioning_activated field and report any errors
+        try:
+            for entry in data["entry"]:
+                if entry["name"] == "general":
+                    return bool(int(entry["content"]["versioning_activated"]))
+        except KeyError as e:
+            raise KeyError(
+                "Cannot retrieve versioning status, unable to versioning status using the expected "
```
pyth0n1c, line 156: Very minor typo here. I assume it is `Unable to determine versioning status` instead of `unable to versioning status`?

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,530 @@
...
+        if not self.is_cms_parser_enabled:
+            raise Exception("Something went wrong, cms_parser is still disabled.")
+
+        self.logger.info(
+            f"[{self.infrastructure.instance_name}] cms_parser successfully toggled to force run"
+        )
+
+    def wait_for_cms_main(self) -> None:
+        """
+        Checks the cms_main index until it has the expected number of events, or it times out.
```
pyth0n1c, line 236: Why not raise an exception after timeout instead of just returning?

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,530 @@
...
+
+                # Get the name of the search in the CMS event and attempt to use pattern matching
+                # to strip the prefix and suffix used for the savedsearches.conf name so we can
+                # compare to the detection
+                cms_entry_name = cms_event["action.correlationsearch.label"]
+                self.logger.info(
+                    f"[{self.infrastructure.instance_name}] {offset}: Matching cms_main entry "
+                    f"'{cms_entry_name}' against detections"
+                )
+                ptrn = re.compile(r"^" + self.global_config.app.label + r" - (?P<stripped_cms_entry_name>.+) - Rule$")
```
pyth0n1c, line 391: I wonder if it's better to go in reverse now that we have the following function to generate a search name: 
https://github.com/splunk/contentctl/blob/98573b0fcc6040bc10046bf444c29d2685545ba6/contentctl/objects/abstract_security_content_objects/detection_abstract.py#L101

This also means if we change the template used in the link above to generate the detection name, we don't have to change it again in this function.

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,530 @@
...
+        :param detection: The matched detection
+        :type detection: :class:`contentctl.objects.detection.Detection`
+
+        :return: The generated exception, or None
+        :rtype: Exception | None
+        """
+        # TODO (PEX-509): validate additional fields between the cms_event and the detection
+
+        cms_uuid = uuid.UUID(cms_event["detection_id"])
+        rule_name_from_detection = f"{self.global_config.app.label} - {detection.name} - Rule"
```
pyth0n1c, line 500: See a similar comment above about how we compose the name of the detection.

contentctl/objects/content_versioning_service.py
cmcginley-splunk: Good catch!

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        # Get the data input entity
+        cms_parser = self.service.input("data/inputs/cms_parser/main")                              # type: ignore
+
+        # Convert the 'disabled' field to an int, then a bool, and then invert to be 'enabled'
+        return not bool(int(cms_parser.content["disabled"]))                                        # type: ignore
+
+    def force_cms_parser(self) -> None:
+        """
+        Force the cms_parser to being it's run being disabling and re-enabling it.
```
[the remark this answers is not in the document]
cmcginley-splunk: Another good catch haha. not sure how that got so garbled lol

contentctl/objects/content_versioning_service.py
cmcginley-splunk: Good question. I was trying to isolate validation into a single function; the ethos of this function is just to wait the acceptable amount of time. The ethos of `validate_against_cms` is to raise any validation errors.

Open to refactoring though

contentctl/objects/content_versioning_service.py
cmcginley-splunk: Ahhhhh I don't think I knew about this function, I'll take a look!

contentctl/objects/content_versioning_service.py
pyth0n1c: Makes sense, resolving this one

contentctl/objects/content_versioning_service.py
pyth0n1c: This one is _FAIRLY_ new. In the future, it might be a nice idea to actually associate the app config with a detection so that we don't have to carry/pass it around, making `action_dot_correlationsearch_dot_label` a cached_property instead of a function that takes an argument.

cmcginley-splunk: Updated this block to use this function for matching instead of regex

contentctl/objects/content_versioning_service.py
cmcginley-splunk: Resolved

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,508 @@
...
+                return self._cms_main_job
+            raise Exception(
+                "Attempting to return a cached job against the cms_main index, but no job has been"
+                " cached yet."
+            )
+
+        # Construct the query looking for CMS events matching the content app name
+        query = (
+            f"search index=cms_main sourcetype=stash_common_detection_model "
+            f'app_name="{self.global_config.app.appid}" | fields {", ".join(self.cms_fields)}'
```
pyth0n1c, line 297: Amazing, as this means we support apps with different appids :) Great idea not to hardcode DA-ESS-ContentUpdate!

contentctl/objects/content_versioning_service.py
```diff
@@ -0,0 +1,508 @@
...
+            return Exception(msg)
+        elif cms_uuid != detection.id:
+            # Compare the UUIDs
+            msg = (
+                f"[{self.infrastructure.instance_name}] [{detection.name}]: UUID in cms_event "
+                f"('{cms_uuid}') does not match UUID in detection ('{detection.id}')"
+            )
+            self.logger.error(msg)
+            return Exception(msg)
+        elif cms_event["version"] != f"{detection.version}.1":
```
pyth0n1c, line 497: Are there any other fields we can/should be validating as well? For example, the search field itself, given that we have seen some consistency issues there in the past? Although we can argue that things which _actually_ impact the search at runtime are caught during a later phase of the integration test.

Or is this an issue of "yes, we could probably check this, but then should we check EVERY field?"

pyth0n1c (approved):
Left some comments, but out of curiosity not because changes are required.
I will do a bit more local testing of these changes then get them merged and do a new contentctl release!

pyth0n1c (approved):
Re-approving after additional runtime testing.  Nice, descriptive errors, and I like that testing bails as SOON as we detect this error - it means we don't waste time on a full integration test (which is costly in terms of time AND resource costs) when we know, early on, that there has been a CMS issue.


These errors were stimulated by running a test (with the latest security_content) and all apps, then rerunning the test after making some changes to YML. Since the changes were made in YML, they _should not_ be reflected in cms_main and SHOULD show as error:
create new detection with new name - NOT FOUND IN CMS MAIN
create new detection with new name - CMS MAIN CONTAINS 1 LESS OBJECT THAN EXPECTED
change the version of an existing detection - WRONG VERSION IN CMS MAIN
change the uuid of existing detection - UUID MISMATCH WARNING


These errors are exactly what we would expect to find. Great work!
image

contentctl/objects/content_versioning_service.py
cmcginley-splunk: So, yes! Hahah I have an open ticket PEX-509 (see comment on line 473) to look into this. I think there's a lot of room to validate more here

cmcginley-splunk:
Thanks for the thorough testing Eric!!

PR #307, opened 2024-10-14
Allow testing with the default or custom_index

ax-hsmith:
Here is a slack convo from splunk-usergroups with some background:
https://splunk-usergroups.slack.com/archives/CDNHXVBGS/p1728921299027159?thread_ts=1728662173.952599&cid=CDNHXVBGS

I think I identified a problem with running `contentctl test` with a `custom_index` defined in the detection yml.

In order to run tests using a `custom_index`, one must build a custom app that has the `custom_index` name pre-defined in the config (`your_app/default/indexes.conf`) and then point to the app in `contentctl.yml`. This will ensure the `custom_index` is available on the Splunk instance to run tests against, however, it is not enough.

The problem I found is with the `configure_hec` function. It creates the HEC that ingests the logs, however, it doesn't give "access" to any custom indexes...so when you run a test with a custom index, `contentctl` tries to add the logs to the `custom_index` and subsequently fails with a 400 because the HEC is not configured to talk to the custom index.

To work around that issue, I created a helper function (`get_all_indexes`) to collect all the index names that exist on the Splunk instance, including ones created by other apps. And then I modified `configure_hec` to call that function and supply a list of all the indexes to the POST that creates the HEC. After doing that, I can run tests using the default index and a custom index of my choosing (assuming that it is already created on the Splunk instance).


pyth0n1c:
This looks like a great PR, thank you @ax-hsmith :) 
I think this will be a definite improvement. 
You do point out a good additional (potential) issue:
 `I can run tests using the default index and a custom index of my choosing (assuming that it is already created on the Splunk instance`
 Notably, the a user can supply ANY string in `custom_index` - not just one that actually exists on the server.  Should we do additional validation around this, for example:

1. During server creation time, iterate over all tests and create all relevant indexes?
2. At search runtime, if a custom_index doesn't exist, throw a special error for that search? (this is closest to what we do now).

I'm open to other options as well. I also don't think we need to hold back THIS PR in order to build in the additional checks listed above.

pyth0n1c:
I have remapped these changes from a PR against `main` to a different branch as we will make a few more changes to it.
Merging this right now and will link to that new PR when it is opened.

PR #309, opened 2024-10-16
All more custom indexes

pyth0n1c:
This PR contains changes from the following PR contributed by community member @ax-hsmith.

The additional change we will make here is that the get_all_indexes function will also include the standard contentctl replay index.

We will also consider comment https://github.com/splunk/contentctl/pull/307#issuecomment-2413877026 for additional changes.

### DIFF
--- a/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
+++ b/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
@@ -169,12 +169,16 @@ class DetectionTestingInfrastructure(BaseModel, abc.ABC):
             pass
 
         try:
+            # Retrieve all available indexes on the splunk instance
+            all_indexes = self.get_all_indexes()
+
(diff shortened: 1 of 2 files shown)


pyth0n1c:
See the following error, which is now generated and dumped into test_results/summary.yml, when trying to replay data into an index that does not exist on the target server. It should be nice and verbose, listing the index you attempted to replay into and all the indexes that DO exist on the server.
This is checked at individual test time, not during validate or initial test set up.
image

ax-hsmith:
Hi @pyth0n1c!

First, thanks for considering my PR for merge. It's truly a pleasure to contribute to such an amazing project!

To answer your question, yes, having the exception/error in the results is a welcomed improvement.

I think it also may be worthwhile to update the README or documentation somewhere to indicate "if you wish to use a `custom_index` during tests, you must include a Splunk app in `contentctl.yml` that has the index pre-defined in it's configuration. For more information refer to https://docs.splunk.com/Documentation/Splunk/9.3.1/Admin/Indexesconf#indexes.conf.example (or something similar). I am happy to help with this if you think it's worthwhile.

PR #311, opened 2024-10-19
add --compare_against flag to release_notes action

patel-bhavin:
adds a flag to release_notes action  `-- compare_against` : so that we can mention another branch name to compare and create notes against. Tested locally and the generation of notes looked good!


pyth0n1c (approved):
There may be future work where we place stricter requirements on what subset of arguments can be provided in terms of old_tag/new_tag and branch names, but this is acceptable for now.

PR #324, opened 2024-11-06
Ensure we print the right field for data_source

ljstella:
`data_source` is distinct and different than `data_sources` - we don't appear to have made that distinct enough in some places so even we have made use of the wrong one in some places.


patel-bhavin:
👍  - Looks good!

patel-bhavin (approved):

pyth0n1c (approved):
approved, thanks for the fix!

PR #332, opened 2024-11-21
Pin Pydantic to Resolve Contentctl Bug

pyth0n1c:
Pydantic 2.10 release causes an exception to be thrown in our codebaee. resolving the exception may take some time, so we pin it for now to a working version.


cmcginley-splunk (approved):
LGTM

PR #335, opened 2024-12-04
Remove use enum values

pyth0n1c:
Remove the use of all `use_enum_values` in Pydantic Models.
Secondly, clean up the use of `.value` when checking/comparing the values of enums in the codebase.


ljstella (approved):
LGTM

PR #336, opened 2024-12-07
Add contentctl_library_version contentctl.yml field

pyth0n1c:
Begin tracking the version of contentctl used to build a specific repo state and enforcing it
so that builds and other operations are repeatable.


josehelps (approved):
LGTM

pyth0n1c:
Closing out this PR.  Instead of building all of this as another check and feature into contentctl, we simply included a requirements.txt in the security_content repo.  Other users of contentctl should take a similar approach for more deterministic builds:
https://github.com/splunk/security_content/blob/develop/requirements.txt

PR #340, opened 2024-12-12
GH Actions Matrix update

ljstella:
Two simple changes here:
1. Added the GitHub Actions ecosystem to our dependabot config. This will give us PRs when there's new versions of actions, which helps avoid some of the "old runtime" warnings we've seen in our other repos.
2. Shrinks the test matrix for `testEndToEnd.yml` and `test_against_escu.yml`. With [this announcement](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/), the Ubuntu 20.04 runner image is being retired in April. This gets it off our plate entirely a bit sooner. I've also removed the second macos image from the matrices, for simplicity and to reduce total time waiting for otherwise-duplicate jobs to finish.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,8 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 6
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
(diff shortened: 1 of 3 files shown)


pyth0n1c (approved):
These all make sense - good to be proactive instead of reactive

PR #344, opened 2025-01-14
add d assistant Proof of Concept - Step 4 - ESCU 5.0

pyth0n1c:
d assistant poc


josehelps:
This should be done as a YAML against on security_content and not embedded into contentctl.

PR #351, opened 2025-01-22
Fixes for 5.0.0a2

ljstella:
### What's included?

- [ ] Last vestiges of detection.tag.observable in testing
- [ ] Field Aliases fix in Risk_Event.py from @cmcginley-splunk 
- [ ] Migrated not-currently-in-use drilldown code to use detection.rba instead of detection.tags.observable
- [ ] Removed commented out code using observables
- [ ] https://github.com/splunk/contentctl/pull/352 
- [ ] transforms.conf formatting for `match_type`
- [ ] hunting searches not having a Risk config caused issues with testing them
- [ ] .mlmodel files renamed in err


ljstella:
Last commit that removed Observables from the DetectionTags model caught these two stragglers: https://github.com/splunk/security_content/pull/3289

patel-bhavin (approved):

PR #353, opened 2025-01-23
Additions to DataSource model

ljstella:
https://github.com/splunk/security_content/pull/3287/ <- PR to add fields to the datasource object.

At the moment, I've typed these **incredibly** naively with `separator_value` as an `Optional[str]`, and `mitre_components` as `list[str]`. I think `separator_value` may stay that way, but there's a chance with `mitre_components` to leverage some of the work we've done elsewhere on MITRE enrichment.


pyth0n1c:
Minor feedback - in line with latest PEP guidance, we should use 
`variable_name: None | TypeA` 
rather than
`variable_name: Optional[None]`

ljstella:
After conversation w/ the folks that caused this PR to kickoff, looks like we should add some validation around the `mitre_component` portion to ensure that its name matches one of the `x-mitre-data-component` entities from ATT&CK. Validating that the field matches should prevent issues with bad data leaking downstream.

ljstella:
After conversation this week, this should now be good to land in association with https://github.com/splunk/security_content/pull/3287

Validations for it can come later when we have a better idea of how we might want to use it (likely on research.splunk.com)

pyth0n1c (approved):
This looks good. I agree with the PR owner that we can build additional validations for these fields, as necessary, when we have some content ready to use it.

PR #355, opened 2025-01-23
D assistant take 2

pyth0n1c:
This contains a bit of restructuring, but mostly the logic to support content lifecycle management.
Most notably, it supports deprecating and eventually removing content in a scalable way.


patel-bhavin:
Hi Eric, a few comments :

- Since we are creating this automatically at build time , should we drop this lookup file in dist/ and throw in a stanza in transforms? We can also maintain this in the [lookups directory](https://github.com/splunk/security_content/pull/3363) as the website will need this file as well. 

- Lets  add the name of the replacement detection name to the CSV (it would better for the dashboard)

patel-bhavin:
Final test output :

image

patel-bhavin (approved):

PR #356, opened 2025-01-24
v4 to v5 Migration guide, wizard updates

ljstella:
Threw together some details around the largest changes to help folks migrate from v4 to v5 of contentctl.

Also merged in some updates to the new content wizard so it functions again.


pyth0n1c:
We should probably add a `status: production` as a default field to all newly created analytic stories.

pyth0n1c (approved):
Text and functionality of contentctl --new looks good here.

PR #358, opened 2025-01-26
 status_aware_description field

pyth0n1c:
This PR adds `status` to a number of different SecurityContentObjects.
This field can be used to track status (production, deprecated, and possibly experimental).
This field has also be used to generate a "description" field possibly with an experimental or
deprecated warning on that content. 
This is a computed field which is part of SecurityContentObject.
Attempting to access this field if status is NOT present in the object will result in 
NotImplementedError.  This was chosen over simply returning the original
description to make failure to implement the status field an obvious error.

These status fields are now REQUIRED for Baselines, Stories, and Investigations.  They were already required for Detections.

Please note that this PR should be tested with the following Security Content PR:
https://github.com/splunk/security_content/pull/3295


ljstella (dismissed):
LGTM

both validate and build ran without issue against the linked content PR

ljstella (approved):
:shipit: 

LGTM

PR #361, opened 2025-02-03
Add output_fields to datasource class

ljstella:
Ref: https://github.com/splunk/security_content/pull/3301/


pyth0n1c:
We still need to determine if, or when, these additional validations will be merged in. Perhaps we should try running this branch, after resolving merge conflicts, with the latest in security_content to see if any errors are thrown and proceed from there (possibly fixing issues before approving this PR)?

PR #362, opened 2025-02-04
deterministic output for rba dict

pyth0n1c:
define serialization functions
for rba and risk bits in line with
normal pydantic workflow.

There is currently an issue where, because RBA objects now exist in a set,
the serialization order when writing to conf files can change.  Because
of that, the `contentctl inspect metadata validation` flow can detect 
non-meaningful changes in a conf file which, nevertheless, require
bumping version of a detection due to conf file changes.


ljstella (approved):
LGTM

PR #365, opened 2025-02-06
Update appinspect flags

pyth0n1c:
Improve/cleanup usage of appinspect api


patel-bhavin:
Tested this change locally:

image

patel-bhavin:
the report looks good : ALL THE TAGS

image

PR #368, opened 2025-02-07
Adding output fields to data_source

ljstella:
Replacing #361 which has taken on a new purpose. 

This is targeting https://github.com/splunk/security_content/pull/3301


pyth0n1c (approved):
This looks good to me. A more straightforward PR like this makes sense to merge immediately rather than a more robust one as we figure out the future usage of the DataSource Object.

PR #371, opened 2025-02-13
Wrong type for lookup.default_match

pyth0n1c:
A bug was reported where `default_match` was typed as a `bool`, but in reality should be a `str`.
This has been fixed and it has been given a default value of `''` in line with the default value of this field per transforms.conf documentation: https://docs.splunk.com/Documentation/Splunk/9.4.0/Admin/Transformsconf#:~:text=default_match%20%3D%20%3Cstring%3E%0A*%20If%20min_matches%20%3E%200%20and%20Splunk%20software%20has%20less%20than%20min_matches%20for%20any%0A%20%20given%20input%2C%20it%20provides%20this%20default_match%20value%20one%20or%20more%20times%20until%20the%0A%20%20min_matches%20threshold%20is%20reached.%0A*%20Default%3A%20empty%20string.

### DIFF
--- a/contentctl/input/director.py
+++ b/contentctl/input/director.py
@@ -1,30 +1,29 @@
 import os
 import sys
-from pathlib import Path
 from dataclasses import dataclass, field
-from pydantic import ValidationError
+from pathlib import Path
 from uuid import UUID
-from contentctl.input.yml_reader import YmlReader
 
-from contentctl.objects.detection import Detection
-from contentctl.objects.story import Story
+from pydantic import ValidationError
 
-from contentctl.objects.baseline import Baseline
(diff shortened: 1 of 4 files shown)


pyth0n1c (commented):
transforms.conf is identical after the change
image

This also fixes an issue where if default_match was not defined, it would be written as "false" in any api objects with a serialized lookup. For instance:
image

Because this is now always treated as a string (not a bool) if this field is not defined then it is an empty string, which is the default value for this field:
https://docs.splunk.com/Documentation/Splunk/9.4.1/Admin/Transformsconf#:~:text=provided%20(see%20below).-,default_match,-%3D%20%3Cstring%3E%0A*%20If%20min_matches

PR #373, opened 2025-02-18
fix dashboard file names

pyth0n1c:
we must go back to using the file path. per appinspect, filenames are not permitted to have spaces.


patel-bhavin (approved):

patel-bhavin:
The build and inspect both checks out fine with this update!

PR #374, opened 2025-02-24
Enable direct attack_data repo download

pyth0n1c:
If attack_data repo exists, then
use it instead. otherwise, or
if the file is not found in the
repo, default to a download


pyth0n1c:
This PR has been superseded by the following:

https://github.com/splunk/contentctl/pull/392

PR #375, opened 2025-02-26
Convert warning for missing datasource to an error

pyth0n1c:
For example, if a datasource is not defined now you will see an error like the following screenshot.
This is similar to how we output any other "missing" or undefined content, like references to macros/lookups/stories that do not exist.
image


patel-bhavin:
works as expected locally :

image

patel-bhavin (approved):

PR #377, opened 2025-03-05
Data source output fields validation

P4T12ICK:
new validation:
- Checks if all output_fields defined in a data_source object is present in the detection

### DIFF
--- a/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
+++ b/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
@@ -89,7 +89,7 @@ class DetectionTestingManagerOutputDto:
     start_time: Union[datetime.datetime, None] = None
     replay_index: str = "contentctl_testing_index"
     replay_host: str = "CONTENTCTL_HOST"
-    timeout_seconds: int = 60
+    timeout_seconds: int = 120
     terminate: bool = False
 
 
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -1059,25 +1059,26 @@ class Detection_Abstract(SecurityContentObject):
     @model_validator(mode="after")
     def validate_data_source_output_fields(self):
         # Skip validation for Hunting and Correlation types, or non-production detections
-        if (self.status != DetectionStatus.production or 
-            self.type in {AnalyticsType.Hunting, AnalyticsType.Correlation}):
+        if self.status != DetectionStatus.production or self.type in {
+            AnalyticsType.Hunting,
+            AnalyticsType.Correlation,
+        }:
             return self
 
         # Validate that all required output fields are present in the search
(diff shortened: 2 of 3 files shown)


contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
```diff
@@ -89,7 +89,7 @@ class DetectionTestingManagerOutputDto:
     start_time: Union[datetime.datetime, None] = None
     replay_index: str = "contentctl_testing_index"
     replay_host: str = "CONTENTCTL_HOST"
-    timeout_seconds: int = 60
+    timeout_seconds: int = 120
```
pyth0n1c, line 92: Did you have some timeout issues when testing that required this bump?

contentctl/objects/data_source.py
```diff
@@ -23,7 +23,7 @@ class DataSource(SecurityContentObject):
     field_mappings: None | list = None
     convert_to_log_source: None | list = None
     example_log: None | str = None
-    output_fields: list[str] = []
+    output_fields: None | list = None
```
pyth0n1c, line 26: We have tried to move away from fields with different types, when possible, because it reduces the checking that needs to be done.
Functionally, an empty list here serves the same purposes as None, doesn’t it?

contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -1055,3 +1055,29 @@ def get_summary(
...
+    def validate_data_source_output_fields(self):
+        # Skip validation for Hunting and Correlation types, or non-production detections
+        if (self.status != DetectionStatus.production or 
+            self.type in {AnalyticsType.Hunting, AnalyticsType.Correlation}):
+            return self
+
+        # Validate that all required output fields are present in the search
+        for data_source in self.data_source_objects:
+            if not data_source.output_fields:
+                continue
```
pyth0n1c, line 1069: If we change the typing of output_fields back to `output_fields: list[str] = []` then we don't even need this check. This list comprehension below will always return [] because there are no output fields.

pyth0n1c (commented):
Just a few comments on the changes

contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructure.py
P4T12ICK: yea for the new detections which uses two tstats searches with prestats

contentctl/objects/data_source.py
P4T12ICK: okay with it. I will change it back

contentctl/objects/abstract_security_content_objects/detection_abstract.py
P4T12ICK: This is still needed even after making the change to the type of the output_fields

pyth0n1c (approved):
Reviewed changes with the author, checked code and tested locally.
Looks good to me, approved!

PR #378, opened 2025-03-05
updated mitre map gen

josehelps:
updates layer to latest also to mitre att&ck v16
image


josehelps:
[coverage.json](https://github.com/user-attachments/files/19096891/coverage.json)
example generated file

pyth0n1c (approved):
I have also reviewed the output file this creates and it looks correct to me

PR #379, opened 2025-03-08
Update RBA API Output

pyth0n1c:
Make sure that relevant/required fields are included in the detections.json API object

### DIFF
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -435,12 +435,8 @@ class Detection_Abstract(SecurityContentObject):
             "datamodel": self.datamodel,
             "source": self.source,
             "nes_fields": self.nes_fields,
+            "rba": self.rba,
         }
-        if self.rba is not None:
-            model["risk_severity"] = self.rba.severity
-            model["tags"]["risk_score"] = self.rba.risk_score
-        else:
-            model["tags"]["risk_score"] = 0
 
         # Only a subset of macro fields are required:
         all_macros: list[dict[str, str | list[str]]] = []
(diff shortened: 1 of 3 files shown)


contentctl/objects/abstract_security_content_objects/detection_abstract.py
```diff
@@ -435,12 +435,8 @@ def serialize_model(self):
             "datamodel": self.datamodel,
             "source": self.source,
             "nes_fields": self.nes_fields,
+            "rba": self.rba,
         }
-        if self.rba is not None:
```
pyth0n1c, line 439: Since these are encoded in individual risk_objects, I do not believe they should be included at a high-level for the entire RBA section anymore.

contentctl/objects/rba.py
```diff
@@ -11,6 +11,23 @@
 RiskScoreValue_Type = Annotated[int, Field(ge=1, le=100)]
 
 
+def risk_score_to_severity(num: int) -> RiskSeverity:
```
pyth0n1c, line 14: If risk_severity needs to exist both for individual risk_objects AND at a high level for the entire RBA section, then this should be broken out here so it can be re-used.
It is also used in the savedsearches.conf dump logic here:
https://github.com/splunk/contentctl/blob/a64d8795ae8f6f796230d397d4f475daf9c55cd0/contentctl/output/templates/savedsearches_detections.j2#L69
Is this a required field that MUST still be present in savedsearches.conf?

pyth0n1c:
After discussing with relevant team(s), this PR will likely be closed out in favor of another that implements the relevant changes.
We will keep this open until that point for documentation purposes.

pyth0n1c:
This PR has been closed. New work will be focused on the following PR, which supersedes this one:
https://github.com/splunk/contentctl/pull/390

PR #387, opened 2025-03-27
Adding risk data model validation to integration testing

cmcginley-splunk:
# Context
* ES 8+ now allows notables (Findings) to have risk data associated with them
* We have TTP detections which generate both risks and notables, so we want to ensure risk data is only being associated with risks in this case to avoid risk score doubling
* We also have Correlation detections which generate notables alone (often based on an aggregation of existing risk entities); as of now, we don't define any risk modifiers for these, so we wouldn't expect them to be associated w/ risk data
* It's possible that ESCU's behavior in the future may change, but as of now, we do not expect ESCU detections to generate notables which appear in the risk data model

# Code changes
* Added new methods to `CorrelationSearch` class which retrieves events from the risk data model
* During notable validation, risk data model is checked to ensure no notables appear in there
* New base class added to support notable and risk events

# Testing
* TODO

# TODOs
* [ ] disable logging
* [ ] address outstanding in-line TODOs


contentctl/objects/correlation_search.py
```diff
@@ -509,6 +513,9 @@ def risk_event_exists(self) -> bool:
         events = self.get_risk_events(force_update=True)
         return len(events) > 0
 
+    # TODO (cmcginley): to minimize number of queries, perhaps filter these events from the
+    #   returned risk dm events? --> I think no; we want to validate product behavior; we should
+    #   instead compare the risk dm and the risk index (maybe...)
     def get_risk_events(self, force_update: bool = False) -> list[RiskEvent]:
```
cmcginley-splunk: There's a possible optimization to be done here, but I fear it may make the code a bit messier.

Currently, we query the risk index directly, the notable index directly and now the risk data model directly. In the case of risk objects, we are effectively querying these twice now as a result (potentially querying notables events twice as well, but that is an error case).

We could cut down on queries by changing `get_risk_events` to invoke `get_risk_dm_events` and filter on the returned set to be just those from the risk index. What do you think @pyth0n1c ? Is the juice worth the squeeze on this optimization?

contentctl/objects/correlation_search.py
```diff
@@ -637,15 +646,116 @@ def get_notable_events(self, force_update: bool = False) -> list[NotableEvent]:
...
+                f"Using cached risk data model events ({len(self._risk_dm_events)} total)."
+            )
+            return self._risk_dm_events
+
+        # TODO (cmcginley): optimize this query? don't REALLY need the full events here for the
+        #   depth of validation we're doing -> really just need the index
+        # TODO (#248): Refactor risk/notable querying to pin to a single savedsearch ID
+        # Search for all risk data model events from a single scheduled search (indicated by
+        # orig_sid)
+        query = (
```
cmcginley-splunk, line 694: As the code is currently written, we don't actually need to fetch the full events; we could just fetch the index field. This would make this query a little less expensive.

Going this route is in compatible with the other optimization proposed earlier (re: having `get_risk_events` call `get_risk_dm_events`). Curious about your thoughts @pyth0n1c

pyth0n1c: While this is indeed a small optimization, I think fetching the full events is more in line with what happens during the normal operation of ES.  While not what this check is about (in a very specific way), giving slightly more opportunities for errors is actually a good thing (even though I would not **expect** this to produce more errors).

Additionally, while we're only using tine `index` field _for now_ I'm not sure what the future will bring.

I CAN see this being a larger issue if we wind up pulling back thousands of Risk events - but that should be rare and we will probably want to improve that search anyway. And even then, that issue is likely "test took an extra few seconds to run" not a failure or unexpected result.

contentctl/objects/correlation_search.py
```diff
@@ -33,7 +34,7 @@
 from contentctl.objects.risk_event import RiskEvent
 
 # Suppress logging by default; enable for local testing
-ENABLE_LOGGING = False
+ENABLE_LOGGING = True
```
pyth0n1c, line 37: I am going to switch this to False.  
But we should probably consider exposing ALL logging and verbosity as a common setting somewhere (possibly in contentctl.yml or the command line) since as we add more and more of it, it can be extremely useful

pyth0n1c (approved):

contentctl/objects/correlation_search.py
pyth0n1c: I think at this time, probably not.  I agree this is trading some complexity for some speed - but in the scheme of things this is not an area that takes a particularly long time in comparison to all the other "stuff" contentctl does. I would err on the side of less complexity in this case - if we wind up improving contentctl speed significantly in the future we can come back and re-evaluate.

pyth0n1c (approved):
I made two additional small changes to bump the version and disable logging.
Given those are small changes, I will approve and merge myself.

PR #388, opened 2025-03-28
Bumping to ruff v0.11.2

ljstella:
No need to merge this urgently, I'll keep bumping it as it gets updated.


pyth0n1c (approved):
Staying on top of even small updates like this to keep our PR backlog as manageable as possible. Looks good.

PR #390, opened 2025-03-29
update detections.json output with new rba structure

pyth0n1c:
Move risk_score and risk_severity from
tags into higher level Detection_Abstract
object. Make sure risk_severity
is consistent and included only in
json content that requires it.
Updated serialization logic for
detections.json

Note that this PR also supercedes the following PR, which has been closed:
https://github.com/splunk/contentctl/pull/379

### DIFF
--- a/contentctl/objects/abstract_security_content_objects/detection_abstract.py
+++ b/contentctl/objects/abstract_security_content_objects/detection_abstract.py
@@ -43,10 +43,11 @@ from contentctl.objects.enums import (
     DetectionStatus,
     NistCategory,
     ProvidingTechnology,
+    RiskSeverity,
 )
 from contentctl.objects.integration_test import IntegrationTest
 from contentctl.objects.manual_test import ManualTest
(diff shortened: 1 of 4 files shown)


pyth0n1c:
I have submitted a draft of the new objects to the relevant team and am waiting on feedback. This PR should remain in DRAFT until we receive that feedback and confirmation that it is correct.

pyth0n1c:
I have taken a look by diffing the old JSON files and old Application files side by side.  The only notable differences are the expected changes around `rba` and related objects in the detections.json files.

Some of the fields in some of these JSON files have their orderings change on subsequent runs - this ordering is not meaningful for our purposes (for example a list of MITRE enrichments) but we may consider sorting them in the future since they make diffing much more challenging and time-consuming.

patel-bhavin:
verified the generation of the output and manually compared with https://securitycontent.scs.splunk.com/detections.json
 
image

The generation json file has been tested by the SSE team.

PR #391, opened 2025-03-30
release notes - add windows path support

0xC0FFEEEE:
Simple change that converts trailing slashes of content directories to blackslashes on Windows systems.

### DIFF
--- a/contentctl/actions/release_notes.py
+++ b/contentctl/actions/release_notes.py
@@ -3,6 +3,7 @@ from git import Repo
 import re
 import yaml
 import pathlib
+from os import path
 from typing import List, Union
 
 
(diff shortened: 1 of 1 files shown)


pyth0n1c (commented):
I think a better way to do this is probably a slightly more complex change, but probably more pythonic? We should consider converting the directories from strings to actual pathlib.Path objects and then using `file_path.is_relative_to(directory_path)`

While this requires more changes in the file, I think this is a lot cleaner and "more correct" - any thoughts on this change instead?

pyth0n1c:
@0xC0FFEEEE I've made the alternative change here which updates the typing on the directory field - can you give a look at that one as well and tell me your thoughts in comparison to this one?
Thanks!

https://github.com/splunk/contentctl/pull/393

pyth0n1c:
I am going to close out this PR and merge the following PR, linked above, which has a slightly different solution implemented by converting a list of directory names as strings to more proper Python pathlib.Path objects.

https://github.com/splunk/contentctl/pull/393

PR #392, opened 2025-04-01
Enable Attack Data Download before Test

pyth0n1c:
If attack_data repo exists, then
use it instead. otherwise, or
if the file is not found in the
repo, default to a download.

If the file is not found, but we SHOULD have found it, we throw an error at validate time.

This PR supersedes the previous attempt at this implementation:
https://github.com/splunk/contentctl/pull/374


ljstella (dismissed):
LGTM

ljstella (approved):
:shipit:

PR #393, opened 2025-04-01
Fix pathing issue on Windows and Cleanup Typing

pyth0n1c:
A PR came in  from @0xC0FFEEEE here with a fix for release_notes on Windows:
https://github.com/splunk/contentctl/pull/391#pullrequestreview-2734080802

Based on my comment, we may want to convert the names of these directories, which are currently a string, to paths so that we can be a bit more pythonic.  
@0xC0FFEEEE , could you give your thoughts on whether this ALSO fixes your issue while improving the code in a different way as well?

Thanks!

### DIFF
--- a/contentctl/actions/release_notes.py
+++ b/contentctl/actions/release_notes.py
@@ -1,9 +1,11 @@
-from contentctl.objects.config import release_notes
-from git import Repo
-import re
-import yaml
 import pathlib
+import re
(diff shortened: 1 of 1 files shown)


0xC0FFEEEE:
Living up to the username @pyth0n1c 😉 

This is much better than my proposed fix!

PR #398, opened 2025-04-18
Fix runtime csv testing error

pyth0n1c:
testing with `contentctl test mode:changes ...` would fail due to the presence of the new RuntimeCSV type.
The code was checking for the filename of a RuntimeCSV Lookup, however this does not exist and is set to None, since it exists only at runtime.

Now, we just exclude instances of RuntimeCSV from the mode:changes checking to see what needs to be tested.


patel-bhavin (approved):

patel-bhavin:
image

creates the plan correctly!

contentctl test --plan-only mode:changes --mode.target-branch develop

PR #403, opened 2025-04-30
New error printout generic case

pyth0n1c:
We were missing printouts of some generic errors, specifically around the tests section.
There isn't much nicer formatting we can apply to this since the message is complex, but we do need to make sure it is emitted.


ljstella (approved):
:shipit:

PR #406, opened 2025-05-06
Risk from str to int

pyth0n1c:
Change the type of risk_score in _entities in the conf file from str to int.
This PR just removes the quotes to change
`"risk_score": "0"`
to
`"risk_score": 0`
At the dev's request


ljstella (approved):
:shipit:

patel-bhavin (approved):

PR #408, opened 2025-05-13
Bump api to v2

pyth0n1c:
Since the structure of these json objects (really only detections.json, but there are some dependencies on this object from the other objects) has changed a bit, we bump the names to reflect that they are updated.


patel-bhavin:
Tested this locally :
image


Generated files with v2 correctly!

image

patel-bhavin (approved):

PR #410, opened 2025-05-19
Catch duplicate analytic stories and other mapped content

pyth0n1c:
This only catches references to other SecurityContentObjects. 
There is still the potential for lots of duplicates in other annotations, like references, MITRE Tags, CVE, etc, but this is still an improvement.


patel-bhavin:
Looking solid on local tests :

image

patel-bhavin (approved):

PR #411, opened 2025-05-28
Bumping Ruff version

ljstella:
Bumping Ruff in pyproject + precommit

Enabling some new rulesets in Ruff, too


pyth0n1c (approved):
Looks good. The vast majority of these changes are straightforward, like improving string formatting in fstrings and cleaning up/sorting imports related to updated ruff config.
There are a few other updates which are also minor which look good.

I do think the updated string formatting will take a moment for many developers who are unfamiliar with this string format to understand - but it is a more modern and pythonic way to do things, so I approve of the change.

PR #415, opened 2025-06-13
Adding new macro & lookup exceptions

ljstella:
https://github.com/splunk/security_content/pull/3544/ depends on the use of the `mitre_attack_lookup` lookup provided by ES.

https://github.com/splunk/security_content/pull/3558 depends on the use of the `globedistance()` macro and the `asn_lookup_by_cidr` lookup, both provided by ES.

This stubs them both into the ignore lists- detections will be updated w/ appropriate flags (We should probably have a different state that allows for unit tests to run when ES is present instead of having to default to manual_testing?)


pyth0n1c (approved):
These changes look good.  In the future, we should come up with a more scalable way to do this type of thing (allowing reference to dependencies in other apps).
Ideally, we should not have to update the contentctl codebase for this, we should be able to update a "content file"

pyth0n1c (approved):
just a version bump this time, merging

PR #421, opened 2025-06-25
Default Stanza to prevent issues post-search-removal

ljstella:
Spoke w/ @pyth0n1c about this approach to prevent catching unwanted defaults from the system level. Creating this as a start.

### DIFF
--- a/contentctl/output/templates/savedsearches_detections.j2
+++ b/contentctl/output/templates/savedsearches_detections.j2
@@ -1,10 +1,13 @@
 ### {{app.label}} DETECTIONS ###
 
+[ default ]
+disabled = 1
+
 {% for detection in objects %}
 [{{ detection.get_conf_stanza_name(app) }}]
 action.escu = 0
 action.escu.enabled = 1
-description = {{ detection.status_aware_description | escapeNewlines() }} 
+description = {{ detection.status_aware_description | escapeNewlines() }}
 action.escu.mappings = {{ detection.mappings | tojson }}
 action.escu.data_models = {{ detection.datamodel | tojson }}
 action.escu.eli5 = {{ detection.status_aware_description | escapeNewlines() }}


contentctl/output/templates/savedsearches_detections.j2
```diff
@@ -1,10 +1,14 @@
 ### {{app.label}} DETECTIONS ###
 
+[ default ]
```
pyth0n1c, line 3: Looks like this MUST be 
`[default]`

as (note the extra spaces)
`[ default ]`

is picked up as a new stanza

pyth0n1c (changes requested):
I believe this MUST (should?) be `[savedsearches/default]`

contentctl/templates/app_template/metadata/default.meta
```diff
@@ -6,18 +6,9 @@ export = system
...
-
-## Postprocess
-[postprocess]
-access = read : [ * ], write : [ * ]
-
+## DO NOT EXPORT THE [default] stanza, and the [default] stanza alone.
+## Because this comes later in the default.meta file, it overrides the
+## export = system for [] above.
+## We MAY want to consider change the access, like making this stanza read-only or similar
+[savedsearches/Default]
```
pyth0n1c: I believe this MUST (certainly should) be `[savedsearches/default]`, right?

cob16:
For future readers upgrading to [v5.5.11](https://github.com/splunk/contentctl/releases/tag/v5.5.11) please note if you are moving to this version you will need to update your app_template the same way as https://github.com/splunk/security_content/pull/3815
Otherwise the defaults will become global and this will disable all non contentctl saved searches in your instance.

@ljstella 
I think it would help if the example change in https://github.com/splunk/security_content/pull/3815 was linked in the [v5.5.11](https://github.com/splunk/contentctl/releases/tag/v5.5.11) release notes. Yes we should have seen this coming when reviewing the changelog but I think this is going to catch quite a few people out that don't realize that app configs can pollute global configuration in this way and adds a bit of a nice warning to help others from making the same mistake in a similar way to https://github.com/splunk/contentctl/pull/472 .

PR #433, opened 2025-08-06
Update to add SPLUNK_GENERALTERMS env. variable

GasPar-Sec:
To deploy a the recent Splunk container, it is required to accept the Splunk General Terms.
This can be done by adding the following environment variable to the docker run command:

* SPLUNK_GENERALTERMS=--accept-sgt-current-at-splunk-com

### DIFF
--- a/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructureContainer.py
+++ b/contentctl/actions/detection_testing/infrastructures/DetectionTestingInfrastructureContainer.py
@@ -90,6 +90,7 @@ class DetectionTestingInfrastructureContainer(DetectionTestingInfrastructure):
         ]
 
         environment = {}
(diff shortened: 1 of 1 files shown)


pyth0n1c:
Thanks for the PR @GasPar-Sec - I need to determine if this causes an issue when starting pre-10.x containers as well.
If it does NOT cause an issue, then I will merge the PR.
It if DOES cause an issue, then I will have to build a bit of logic to ONLY include this parameter on 10.x+ instances.

For other folks looking at this PR, here is some documentation around this new, required flag:
https://hub.docker.com/r/splunk/splunk#:~:text=Starting%20in%2010.x%20image%20versions%20of%20Splunk%20Enterprise%20and%20Splunk%20Universal%20Forwarder%2C

PR #442, opened 2025-10-07
π-thon support

ljstella:
Python 3.14 [has been released](https://docs.python.org/3.14/whatsnew/3.14.html). Temporarily opening this PR to run tests. 

Also temporarily removed 3.11 from the matrix to not add too many tests.

### DIFF
--- a/contentctl/objects/config.py
+++ b/contentctl/objects/config.py
@@ -77,14 +77,32 @@ class App_Base(BaseModel, ABC):
 
 class TestApp(App_Base):
     model_config = ConfigDict(validate_default=True, arbitrary_types_allowed=True)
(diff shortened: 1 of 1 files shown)


ljstella:
It _appears_ as if python 3.14 support is going to be blocked by a pydantic update that is blocked the current usage pattern of tyro. 

cc: @pyth0n1c given this likely applies to other related tooling.

ljstella:
Okay - looks like dependency wise, we're mostly squared away these days. Probably can tighten up the pydantic & tyro version restraints. But we're now at a point where we get an identical error in both old versions and 3.14, related to actual changes tyro has made.

PR #449, opened 2025-10-23
PEX-653: KVStore Versioning Validation

xqi-splunk:
### Overview

This PR make below changes:

1. Add kvstore versioning validation while preserving previous datastore based versioning validation logic
2. Add new wait versioning activation logic for kvstore since in ES 8.3.0 versioning is activated by default but would need to wait until is fully activated. Using new API endpoint to check versioning activation.
3. Create a CMSEvent class for cms_event

### Testing

1. Tested with ES 8.3.0: 
Screenshot 2025-10-23 at 1 03 52 PM
2. Tested with ES 8.0.2:
Screenshot 2025-10-23 at 1 00 43 PM

_Note: The reason doesn't test with ES version between 8.0.4 and 8.2.3 is there is known issue with version validation in these versions_

### DIFF
--- a/contentctl/objects/content_versioning_service.py
+++ b/contentctl/objects/content_versioning_service.py
@@ -161,13 +161,13 @@ class ContentVersioningService(BaseModel):
         return es_version is not None and es_version >= Version("8.3.0")
 
     @cached_property
-    def datastore_content_versioning(self) -> bool:
+    def indexbased_content_versioning(self) -> bool:
         """
-        Indicates whether we should test content versioning based on datastore logic. Content versioning
-        should be tested with datastore logic when ES is less than version 8.3.0 but greater than or equal
+        Indicates whether we should test content versioning based on indexbased logic. Content versioning
+        should be tested with indexbased logic when ES is less than version 8.3.0 but greater than or equal
         to version 8.0.0.
 
-        :return: a bool indicating whether we should test content versioning with datastore logic
+        :return: a bool indicating whether we should test content versioning with indexbased logic
         :rtype: bool
         """
         es_version = self.es_version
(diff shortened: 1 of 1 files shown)


contentctl/objects/content_versioning_service.py
```diff
@@ -292,16 +409,25 @@ def _query_cms_main(self, use_cache: bool = False) -> splunklib.Job:
...
-        query = (
-            f"search index=cms_main sourcetype=stash_common_detection_model "
-            f'app_name="{self.global_config.app.appid}" | fields {", ".join(self.cms_fields)}'
-        )
+        if self.kvstore_content_versioning:
+            query = (
+                f"| inputlookup cms_content_lookup | search app_name={self.global_config.app.appid}"
+                f"| fields content"
+            )
+        elif self.datastore_content_versioning:
```
cmcginley-splunk: Why do you use the name `datastore` to describe the index-backed versioning?

xqi-splunk: I just revisited these three terms a little bit. The `datastore` sounds more like a general term that refers to any storage (including kvstore). Would you like me to update all the `datastore` to `indexbased` or `indexbacked`?

cmcginley-splunk: I believe the term datastroe is being used generically here. If you look at the design proposal, they use the language of index-backed and kvstore-backed. We should use the same language here

https://docs.google.com/document/d/1ejweYPz31stX7FjCM5A8_1HS0Wh372qpFsl_uBx4ueo/edit?tab=t.0

cmcginley-splunk (approved):
Approve. We will follow up on the open thread later

PR #472, opened 2026-02-18
Adding warning about default.meta

ljstella:
Put in the savedsearches.conf that you shouldn't copy the `[default]` stanza withou the `default.meta` changes, lest you pollute your whole system with these settings.


ljstella:
Ran `contentctl inspect` against the current state of security_content's develop branch and confirmed that this change did not break detection metadata checking:

Screenshot 2026-02-18 at 11 34 27 AM

0xC0FFEEEE:
@ljstella Tracking down why ES Mission Control not loading today because of this snippet in our newly installed release was fun! Glad you guys spotted it.

What changes are required to default.meta?

0xC0FFEEEE:
nvm, I see now 👍🏼 - contentctl/templates/app_template/metadata/default.meta

ljstella:
@0xC0FFEEEE the version of ESCU that went out with the default stanza had this new `default.meta` file: https://github.com/splunk/security_content/blob/develop/app_template/metadata/default.meta

In the same PR that we made the changes to add the `[default]` stanza, we also added that change to the `app_template` used to init new apps. However, that's not kept in-sync when contentctl is updated. As always, we recommend pinning a version, and then reviewing the full set of changes between versions. 

We're no longer accepting outside contributions on this repo, so continued use of the tool is going to potentially become difficult without a conscientious effort to review the changes that are made.

ThomasNicholson-ho:
The default disabled change just tripped us up too. Appreciate the warning for future users that copy app_template but it could have been better communicated as a breaking change for users with an existing app_template.

PR #473, opened 2026-02-25
Date cannot be future

pyth0n1c:
Generate an error during content validation if the date is set for the future.
Note that this is relative to UTC, so there are some minor edge cases where this generates an exception based on the locale of a developer.  This is documented in the error message returned however.

Note this is what the new check shows on our repo today, in which a piece of content accidentally has a future date:
image


ljstella (approved):
LGTM

pyth0n1c:
Note - current failures against production branch of security_content are to be expected until a fix is merged into that repo.
Intentionally merging this first.

PR #483, opened 2026-09-16
Update all dependencies to latest versions

pyth0n1c:
## Summary
Bumped every dependency in `pyproject.toml` to its latest compatible version.

**Bumped to latest:**
- pydantic, PyYAML, xmltodict, attackcti, questionary, docker, tqdm, pygit2, tyro, gitpython, rich
- dev dependency: ruff

**Judgment calls (not bumped to the absolute latest release):**
- `splunk-sdk` kept at 2.1.1 (latest is 3.0.1, but that requires Python >=3.13, outside this project's supported `^3.11,<3.14` range)
- `setuptools` floor raised to 80.10.2 but kept below 81 (versions >=81 drop `pkg_resources`, which `pycvesearch` imports directly — confirmed this breaks import)
- `tyro` bumped to 1.0.16 after confirming the previously-documented parsing issue with large file counts under `--mode.files` (>130 files) no longer reproduces (tested with 400 files)

**Side effects fixed:**
- The ruff version bump surfaced 19 new lint findings (`RUF036`, `None`-union ordering) and one formatting diff in previously-clean code; fixed both so CI lint/format checks stay green. No behavior changes, purely cosmetic reordering/formatting.

## Test plan
- [x] `poetry lock` / `poetry install` resolve cleanly
- [x] `contentctl --help` runs
- [x] `pkg_resources` / `pycvesearch` import successfully under the pinned setuptools version
- [x] `ruff check` and `ruff format --check` pass
- [ ] CI checks (lint/format, escu smoketest, end-to-end tests) pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pyth0n1c:
This PR updates a significant number of dependencies, but extremely limited testing has been performed on these updates.  Because contentctl is no longer actively maintained, this is a sufficient level of testing to evaluate this PR.
Because two specific library updates failed testing, those dependencies (`pydantic` and `attackcti`) received an explicit comment stating why they were not updated.

As such:
1. I am approving and merging this myself.
2. We are not cutting a new release or version update of contentctl since it is no longer actively maintained.

